Issue #52, opened 2019-08-31
Dismiss is missing


moa77:
For asynch request that needs to be cancelled there is still a need for a dismiss/cancel operation. An idea is to use, as wfs 3, the http delete method.

PR #67, opened 2020-04-30
Tentative fix for #52

gfenoy:
This is a tentative fix for #52 that brings back the [http://docs.opengeospatial.org/is/14-065/14-065.html#86 Dismiss Extension] from WPS 2.0.


core/clause_7_core.adoc
```diff
@@ -200,11 +200,11 @@ The following section describes a method to retrieve the available processes off
 
 ==== Operation
 
-include::requirements/core/REQ_process-collection-op.adoc[]
```
bpross-52n, line 203: I don't understand this change.

gfenoy: Maybe it is a mistake but, I have noticed when trying to produce the output document using asciidoctor that the file was not named REQ_process-collection-op.adoc but REQ_job-collection-op.adoc, hence I have made the modification to include the correct filename.

bpross-52n: Right, the mistake is in clause_10_job_collection.adoc. The software I am using seems to be a bit more forgiving regarding wrong includes..

bpross-52n:
> Thank you very much for this, Gérald. I would have a couple of questions and I think the dismiss operation should be defined as separate requirements class (like Job collection). I am happy to submit a full review for you to update this PR. Alternatively I could take over most of the changes and add them to master, giving you full credit in the documents history. However, that would mean to close this PR eventually and also opengeospatial/master and your master would be out of sync then.. What do you say?

@gfenoy what is your opinion about this?

gfenoy:
> > Thank you very much for this, Gérald. I would have a couple of questions and I think the dismiss operation should be defined as separate requirements class (like Job collection). I am happy to submit a full review for you to update this PR. Alternatively I could take over most of the changes and add them to master, giving you full credit in the documents history. However, that would mean to close this PR eventually and also opengeospatial/master and your master would be out of sync then.. What do you say?
> 
> @gfenoy what is your opinion about this?

Sorry, got issue with internet yesterday. I do agree that making the dismiss operation a conformance class make sense. In addition, Dismiss was an extension in the WPS 2.0.

You are right about closing this PR, I don't have any issue with this. On the other hand, if it helps I can start from your current master then make the dismiss description a conformance class.

Anyway, I think you should choose whatever suits you the best and make your work the easiest possible. I also understand that since May 11th it may be short to handle this. Let me know what you decide please.

Thanks for handling all this work @bpross-52n!

bpross-52n:
Closed via PR #68

PR #85, opened 2020-07-09
fix ref to Execute results/exceptions

tomkralidis:
Fixes dereferencing for synchronous Execute requests.


tomkralidis:
cc @bpross-52n @ghobona

PR #106, opened 2020-11-13
Add optional format to output info

bpross-52n:


bpross-52n:
This PR was accepted in the SWG telecon from November 16th,

PR #107, opened 2020-11-13
Change request/49

bpross-52n:


spacebel:
Finally, I like this change but:
- It doesn't go in the sense of the discussion (https://github.com/opengeospatial/wps-rest-binding/issues/22)
- This changes the default behaviour of previous versions (default = 1)
- Anyway, I think the declaration of types should be more  in line with OpenAPI schema (using array type in case of arrays, as mentionned in https://github.com/opengeospatial/wps-rest-binding/issues/105 )

bpross-52n:
Good point about the default value..

bpross-52n:
This PR was rejected in the SWG telecon from November 16th,

PR #111, opened 2020-11-27
Fix outputInfo

gfenoy:
Fix the reference in the outputInfo.yaml file.


bpross-52n:
Approved in SWG telecon from November 30st

PR #116, opened 2020-12-14
Use inlineOrRefData rather than valueType

gfenoy:
Avoid using inlineValue / reference in outputs as for inputs

### DIFF
--- a/core/openapi/responses/ExecuteSync.yaml
+++ b/core/openapi/responses/ExecuteSync.yaml
@@ -3,8 +3,8 @@ content:
   /*:
     schema:
       oneOf:
-        - $ref: "https://raw.githubusercontent.com/opengeospatial/wps-rest-binding/master/core/openapi/schemas/result.yaml"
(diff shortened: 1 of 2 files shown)


bpross-52n:
Thanks Gérald. We decided on some changes to the schema during the SWG session on Monday, December 7th. I will send around a summary of the changes soon and I will check how this fits in there.

gfenoy:
Thanks a lot for your answer, sorry for missing the meeting.

I would appreciate to know if a decision has been taken about moving to JSON object for input rather than the array currently used. Even in that case, the situation would be the same IMO, meaning that we should consider having the same syntaxe for both input and output (using JSON object).

bpross-52n:
OutputInfo.yaml has been removed. The content moved to [results.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/core/openapi/schemas/result.yaml). @gfenoy Please check if the changes of this PR are reflected there now.

gfenoy:
They are reflected, @bpross-52n thanks for the work.

PR #121, opened 2021-02-01
Adjust URLs for /jobs endpoint

bpross-52n:
As per SWG decision from December 7th, the endpoint for /jobs has been moved to the root level.


bpross-52n:
This is outdated. I will create an updated PR.

PR #141, opened 2021-03-08
Issue 122 - Use JSON Schema fragments to define process inputs and outputs.

pvretano:
This pull request changes the way input and output are described in a process description.  Specifically, the change uses JSON-schema to defined the schema of a process input or output.  As a result, a lot of the existing structures, complexData, literalData, bboxData, etc. can be discarded since all of this can be described using JSON schema.


core/openapi/schemas/inputDescription.yaml
```diff
@@ -1,19 +1,20 @@
...
+    required: 
+      - schema
     properties:
-      input:
-        oneOf:
-          - $ref: "complexDataType.yaml"
-          - $ref: "literalDataType.yaml"
-          - $ref: "boundingBoxDataType.yaml"
+      schema:
+         type: object
```
fmigneault, line 8: I feel that even though the definition of the schema is provided via `schema` property rather than directly inline-ing the field names (`minOccurs`, `maxOccurs`, etc.), there should still be some form of `oneOf` specified here. 

By fully opening the spec this way, there will not be any standard definitions of `BoundingBox` or file reference format for example. I feel this will lead to poor quality in servers implementing OGC-API and very low interoperability between them since it will essentially be free-for-all, and realistically most implementations will not support the full JSON spec for any type of schema.

pvretano: @fmigneault I am not done with the pull request yet.  I had to do an intermediate push so that I had something to present at the last SWG meeting.  So, for a little while longer things will be a little inconsistent.  My next (and hopefully final) push should make everything consistent and that include removing a bunch of definitions (like literalDataType) that are not longer necessary.  Bounding box, however, I think is different.  It should be standardised and its definition should match what is in common, features, records, etc.

PR #172, opened 2021-04-19
Fully qualify the "schema" key so that it is not just a generic object but is in fact a JSON-Schema object (Issue 170).

pvretano:
Get rid of minOccurs and maxOccurs and rely instead on JSON Schema arrays to define process input cardinality.
Define several levels of JSON Schema conformance ranging from very simple to full JSON Schema to allow servers to implement the level on schema complexity that is suitable for their requirements.


core/openapi/schemas/schemaLevel2.yaml
```diff
@@ -0,0 +1,102 @@
...
+        type: integer
+        minimum: 0
+        default: 0
+      pattern:
+        type: string
+        format: regex
+      uniqueItems:
+        type: boolean
+        default: false
+      maxProperties:
```
bpross-52n, line 46: seems to be duplicate, see line 22 of this file. Also minProperties.

core/openapi/schemas/schemaLevel3.yaml
```diff
@@ -0,0 +1,132 @@
...
+      deprecated:
+        type: boolean
+        default: false
+      contentMediaType:
+        type: string
+      contentEncoding:
+        type: string
+      contentSchema:
+        $ref: "schemaLevel3.yaml"
+    patternProperties:
```
bpross-52n, line 130: Keep in mind: If we use this, we will loose swaggerhub capabilities.

spacebel: What are you referring to for this swaggerhub issue ?
Thanks.

bpross-52n (commented):
I am referring to patternProperties. This works only foor OpenAPI 3.1, which swaggerhub does not seem to support yet. I usually compile all the schemas into one OpenAPI document on swaggerhub to do sanity checks and to generate code.

core/openapi/schemas/schemaLevel3.yaml
bpross-52n: I am referring to patternProperties. This works only foor OpenAPI 3.1, which swaggerhub does not seem to support yet. I usually compile all the schemas into one OpenAPI document on swaggerhub to do sanity checks and to generate code.

core/openapi/schemas/reference.yaml
```diff
@@ -0,0 +1,7 @@
+type: object
+required:
+- "$ref"
+patternProperties:
```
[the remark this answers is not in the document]
spacebel: Personnally I replaced all patternProperties by additionalProperties. I think it is critical to keep support on swaggerhub. Moreover, 3.1 spec use another JSON Schema version: this is not a minor change !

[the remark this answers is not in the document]
pvretano: I copied these schema from the files @christophenoel referenced in this issues and I cleaned them up a bit.  The patternProperties key was in the full schema file.  I am not sure how I would replace that with additionalProperties since the patternProperties seems to be enforcing a rule that says that additional properties start with "x-".  So questions:
1. Do we need to enforce this "x-" rule for additional properties ?  If not, we can simple remove the patternProperties key.
2. If we want to keep the "x-" rules, then how would I replace patternProperties with additionalProperties and still enforce that additional properties must start with "x-"?
My vote would be to simply get rid of it since it seems that the "x-" rule is meant to allow the set of JSON schema keys to be arbitrarily extended with additional keys.  I doubt that anyone defining a process input or output would want to do that.

core/clause_7b_jsonschema.adoc
```diff
@@ -0,0 +1,60 @@
...
+JSON Schema Level 1 supports all the capabilities of <<ogc_jsonschema_level0,level 0>> and adds support for objects.
+
+include::requirements/ogc-process-description/jsonschema/REQ_jsonschema_level1.adoc[]
+
+[[ogc_jsonschema_level2]]
+== Requirements Class "JSON Schema Level 2"
+
+include::requirements/requirements_class_jsonschema_level2.adoc[]
+
+JSON Schema Level 2 supports all the capabilities of <<ogc_jsonschema_level2,level 0>> and <<ogc_jsonschema_level1,level 1>> and adds support of an expanded set of schema facets including:
```
fmigneault, line 42: ``<<ogc_jsonschema_level2,level 0>>` should be `<<ogc_jsonschema_level0,level 0>>` ?

PR #174, opened 2021-04-19
Get rid of the mode key.  In all cases the server decides whether to

pvretano:
execute a process synchronously or asynchronously.  The client, however,
can provide hints via the `Prefer` HTTP header defined in RFC 7240.


jerstlouis:
One suggestion was a query parameter for a client to force a particular mode, as opposed to simply indicating a preference, e.g. `respond=sync` or `respond=async`.

One of the advantage of that is that the same execute request (not hard-coding mode) could be used with a query parameter and/or with the prefer header (much like the `f=` query parameter being able to override the `Accept` header).

bpross-52n:
Closed by SWG decision (details will follow)

PR #193, opened 2021-05-26
Return of the prefer header

pvretano:
This is the initial checkin of a PR that replaced the use of the "mode" parameter with the HTTP Prefer header.  This PR needs a bit more work especially in light of the PR #188.


pvretano:
Closing as this change was already merged as part of PR #194.

PR #194, opened 2021-05-28
Prefer header

pvretano:
Remove the mode parameter and instead rely on the HTTP Prefer header and defined default execution mode behaviour.


core/abstract_tests/core/ATS_job-results-success-async-raw-value-multi.adoc
```diff
@@ -1,11 +1,22 @@
...
+
+NOTE: In the case where a process supports both `async-execute` and `sync-execute` job control options there is a possibility that the server responds synchronously even though the `Prefer` headers asserts a `respond-async` preference.  In this case, the following additional test should be performed.
+
+[width="90%",cols="2,6a"]
+|====
+^|Test Method |. Inspect the headers of the synchronous response and see if a `Link` header is included with `rel=monitor`.
+. If the link exists, get the job status as per test <<ats_core_job-op,/conf/cor e/job-op>> and ensure that the job status is set to `successful`.
+. Get the results as per test <<ats_core_job-results-op,/conf/core/job-results>> and verify that they conform to the test <<ats_core_job-results-async-raw-value-multi,/conf/core/job-results-async-raw-value-multi>>.
+. If the link does not exist then verify that the synchronous response conforms to requirement <<req_core_job-creation-sync-raw-value-multi,/req/core/job-creation-sync-raw-value-multi>>.
+|====
```
fmigneault, line 22: Maybe we could consider adding the `Preference-Applied` ([RFC-7240](https://datatracker.ietf.org/doc/html/rfc7240#section-3)) header to be more specific about the applied execution mode based on `Prefer`?

pvretano: Great timing!  I was just about to merge.  I'll add a recommendation about Preference-Applied and then merge!

PR #202, opened 2021-06-07
Small fixes

gfenoy:
Small fixes that should solve various validation issues encoutred when using the openapi-cli and, spectral tools. Some files has been added also to get the swagger-ui required files available.

### DIFF
--- a/core/examples/json/ProcessDescription.json
+++ b/core/examples/json/ProcessDescription.json
@@ -122,7 +122,7 @@
             "contentSchema": "http://schemas.opengis.net/gml/3.2.1/geometryBasic2d.xsd"
           },
           {
-            "$ref": "http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/schemas/geometryGeoJSON.json"
+            "$ref": "http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/schemas/geometryGeoJSON.yaml"
           }
         ]
       }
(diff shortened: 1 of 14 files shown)


pvretano:
@gfenoy we need to discuss some of these changes before I merge.  For example, I don't tjhink that the schemas inputDescription.yaml, outputDescription.yaml or process.yaml are required in the swagger directory.  I'll get in touch with you and we can go over the changes.

gfenoy:
> @gfenoy we need to discuss some of these changes before I merge. For example, I don't tjhink that the schemas inputDescription.yaml, outputDescription.yaml or process.yaml are required in the swagger directory. I'll get in touch with you and we can go over the changes.

It seems easy to justify the existence of every single files listed:
 * process.yaml: it is referenced from the current [ProcessDescription.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/core/openapi/responses/swagger/ProcessDescription.yaml) file,
 * except the fact that indentation in process.yaml is wrong, the file is referencing both outputDescription.yaml and, inputDescription.yaml files, in consequence there are both also required.

Note that both process.yaml, inputDescription.yaml and, outputDescription.yaml are the exact same files as the one available in the schemas directory except that they refer to `..` for files in schemas rather than in the swagger sub-directory.

The ProcessDescription.yaml file is the entry point as it should be referenced from the API published with the OGC API - Processes server instance, for the path: `/processes/{processID}` and response 200.

Another solution would be to create a symbolic link from the swagger sub-directory to schemas yaml files, but I wonder what can be the benefit of doing so.

It will be a pleasure to discuss the matter with you as proposed.

PR #220, opened 2021-06-11
Update OAPIP-specific link relations to be URIs to be compliant with OGC-NA policies.

pvretano:
Update all OAPIR-specific link relations to use the pattern `http://www.opengis.net/def/rel/ogc/1.0/{rel}`.  Eventually these will be
registered with the OGC-NA.


fmigneault:
I just came across the following. 

> item | The target IRI points to a resource that is a member of the collection represented by the context IRI.
https://www.iana.org/assignments/link-relations/link-relations.xhtml

>By default, the context of a link conveyed in the Link header field
   is the URL of the representation it is associated with, as defined in
   [RFC7231], Section 3.1.4.1, and is serialised as a URI.
https://datatracker.ietf.org/doc/html/rfc8288#section-3.2

If I understand those definitions correctly, that means that `self` should always be the current response body, and not individual sub-entries in the collection. For processes in ProcessList, it means `item` should be used instead. 

If this is actually the case, it means that all OGC-API using `self` within a collection listing would also be using the wrong link relation name.

Issue #221, opened 2021-06-11
'dismiss' is not defined in the jobControlOptions


jonherrmann:
Please see:
https://github.com/opengeospatial/ogcapi-processes/blob/master/core/openapi/schemas/jobControlOptions.yaml 

and the example: https://github.com/opengeospatial/ogcapi-processes/blob/master/core/examples/json/ProcessList.json

Discussed in the T17 CITE meeting with Peter and Gérald.

pvretano:
dismiss is already in jobControlOptions.yaml in the master so we can close.

PR #222, opened 2021-06-11
Issue 221

gfenoy:
it should solve the issue #221 

As the server may not implement the dismiss conformance class, I guess there is no need to add the `"dismiss"` item in the `jobControlOption` array.

### DIFF
--- a/core/examples/json/ProcessList.json
+++ b/core/examples/json/ProcessList.json
@@ -1,17 +1,17 @@
 [
     {
-		"id": "EchoProcess",
-		"title": "EchoProcess",
-		"version": "1.0.0",
-		"jobControlOptions": ["async-execute", "sync-execute"],
-		"outputTransmission": ["value",	"reference"],
+	"id": "EchoProcess",
+	"title": "EchoProcess",
+	"version": "1.0.0",
+	"jobControlOptions": ["sync-execute", "async-execute"],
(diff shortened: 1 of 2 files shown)


jerstlouis:
@pvretano @gfenoy 

I'm not sure it is correct to have a "default" in _jobControlOptions.yaml_.

This particular schema only appears in the process description's list of available _jobControlOptions_ as an array, where there really is no default -- every supported control option must be explicitly listed.

There also isn't really a default execution mode (and this is more than about execution mode too -- dismiss is not an execution mode).
Sync execution is only default if the server does advertise `sync-execute` and no _Prefer:_ header is used.

pvretano:
@jerstlouis that was already there so we should NOT mix issues this way.  Create a NEW issue for the default value of `jobControlOptions` and I will create a new PR to fix that.
@gfenoy Please undo the change that changes `process-desc` to `self`.  Again, you are mixing issues and this will cause a merge conflict with the other PR (#220) which was, in fact, about changing `process-desc` to `self`.

jerstlouis:
@pvretano Sorry. Here: #224

bpross-52n:
We need to change process-desc to self, I will take care of it.

bpross-52n:
Merged using command line, see 89389d1f5b1c2054684de640f1f73f0c6b4c8bc5

PR #227, opened 2021-06-14
Handling a mix of value and reference transmissions modes with raw output.

pvretano:
Add requirements and abstract tests to handle the case where the negotiated execution mode is sync or async, the requested response is raw, more that 1 output is requested and a mix of transmission modes (value or reference) are requested.


bpross-52n:
Waiting for Geralds review, SWG agreed to merge.

PR #234, opened 2021-06-18
Oppsie!  Forgot to make processID mandatory if the server supports the Job Lists conformance class.

pvretano:
If the server supports the Job Lists conformance class then the processID property should be mandatory.  Forgot to make this change when I was adding the query parameters to the `/jobs` endpoint.  Fixed now.

### DIFF
--- a/core/abstract_tests/ATS_class_job-list.adoc
+++ b/core/abstract_tests/ATS_class_job-list.adoc
@@ -27,6 +27,8 @@ include::job-list/ATS_jl-links.adoc[]
 
 include::job-list/ATS_type-response.adoc[]
 
+include::job-list/ATS_processid-mandatory.adoc[]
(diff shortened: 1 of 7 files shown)


gfenoy:
It sounds good to me. 

Nevertheless, I don't understand why do we have the following in ats_job-list_processid-mandatory:

"and append the `datatime` query parameter to the request"

I suppose that here, requesting any jobs list with or without parameter should work. Maybe there is a reason I cannot see from the PR itself.

pvretano:
Doh!  @gfenoy typo - cut-paste-error!  Fixed.

gfenoy (approved):
That sounds good to me.

bpross-52n (approved):
If Gerald is happy, I am happy.

PR #242, opened 2021-08-09
Adding status.yaml in schemas

gfenoy:
The `schemas/status.yaml` file has been added to be used as reference from other definitions such as `statusInfo.yaml` and `parameters/status.yaml`.


bpross-52n (changes requested):
Rename status.yaml to statusCode.yaml and update references.

Issue #243, opened 2021-08-09
Colliding names


gfenoy:
Trying to clone the repository, I faced the following warning messages:

````warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same 
colliding group is in the working tree: 

'core/openapi/parameters/processId.yaml' 
'core/openapi/parameters/processid.yaml'
````

As one is processId is used in query and the other one in the path, maybe we should rename the files this way:
 * processIdQueryParam.yaml
 * processIdPathParam.yaml

gfenoy:
Thanks for merge @bpross-52n!

PR #244, opened 2021-08-09
Renaming to avoid collisions 

gfenoy:
Rename processId.yaml and processid.yaml to processIdPathParam.yaml and processIdQueryParam.yaml respectively.

Should solve #243


gfenoy:
Just added a date-time format that was missing in the schema for the datetime parameter definition

PR #260, opened 2021-10-28
Small fixes

gfenoy:
Add the required `type` key in statusInfo.yaml, simplify results.yaml and set nullable to false in statusCode.yaml.


gfenoy:
The [Abstract Test 74 /conf/job-list/processID-mandatory](https://docs.ogc.org/DRAFTS/18-062.html#ats_job-list_processid-mandatory) is explicit about the fact that the `processID` is a required field:

>  Validate that the `processID` property is present in every job.

PR #262, opened 2021-11-08
Correct schema for processIdPathParam

gfenoy:
Here is the correct definition of the processPathParam.


gfenoy:
Actually, this PR simply get back the file back as it was in this [PR](https://github.com/opengeospatial/ogcapi-processes/pull/244).

Issue #283, opened 2022-03-03
AT 16.2, 19 targeting Client, not server; Job Creation -> Process Execution


jerstlouis:
If I understand correctly, the target of the _Processes_ specification is the server (not the client).

Yet some Abstract tests seem to target the client:

A.16. point 2:

> Validate the contents of the POST request using the test [/conf/core/job-creation-request](https://docs.ogc.org/is/18-062r2/18-062r2.html#ats_core_job-creation-request).

Also in A.16 point 3 the links for `/req/core/job-creation-default-execution-mode` and `/req/core/job-creation-auto-execution-mode` are broken.

The whole requirement A.19:

> "Validate that the body of a job creation operation complies with the required structure and contents."

I also think it would be better to use the "process execution" terminlogy instead of "job creation" to reflect the fact that a job is not necessarily created when executing synchronously.

pvretano:
@jerstlouis create a PR so that the SWG can consider the change.

B.T.W. Validating the content of a POST request is not a client requirements.  The server needs not read and validate the content of a POST request ... no?

jerstlouis:
@pvretano These are the abstract tests. My feeling now is that these intended to be talking about the responses to these operations.

pvretano:
@jerstlouis not at all.  Validating the input against a schema that is normatively published in the specification is a valid abstract test.

jerstlouis:
@pvretano The way Abstract test A16.2 and A19 are written right now, it reads as if the request being **sent** to the server must be must be validated. But the test implementation (e.g. the ETS) is what is making that request.

That is why I think was is intended is to validate the **response** from the server to those operations.

So instead of "the contents of the POST request", it should say "the contents of the response to the POST operation".
And instead of "the body of a job creation operation", it should say "the body of the response to the process execution operation".

I also think that this section of the abstract test (A.2.6.1. Job Creation -- AT 16-34) is outdated compared to the requirements, because all of its links to requirements are broken. The requirements might have renamed "Job creation" to "Process execution" since those tests were written. The requirements start with `REQ_process-execute` with anchors starting with `req_core_process-execute`, whereas the ATs are named `ATS_job-creation` and still refer to `req_core_job-creation`.

pvretano:
@jerstlouis create a PR to repair the tests as you see fit.  The SWG can then consider your PR and apply a fix if necessary.

jerstlouis:
[...]

PR #284, opened 2022-03-04
Partially addressing issue #283

jerstlouis:


fmigneault:
I think it is important to preserve the job creation. While execution is more relevant during `sync`, a job is created with status location in `async`. Job also refers to the `/jobs` endpoints. These definitions are what highlights the connection between processes and jobs endpoints.

jerstlouis:
@fmigneault This is only editorial fixes. Sync execution can still result in job creation if the server wishes so, but they are not required. This PR only fixes the abstract tests to match changes already made to the requirements.

fmigneault:
Not only editorial. It modifies items such as `/req/core/job-creation-op`, which are referred by existing APIs in `conformsTo`. 
These should remain as is. I don't mind about the text and abstract tests portions.

jerstlouis:
@fmigneault That change was done prior to 1.0, just the Abstract Tests had not been updated.

fmigneault:
I disagree, they are already found here: https://docs.ogc.org/is/18-062r2/18-062r2.html
This is approved and published for over half a year.

jerstlouis:
@fmigneault In the abstract test. This is also note what goes in `conformsTo`, the list of conformance classes is in [Table 8](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc9).

fmigneault:
No. I am referring to the requirements themselves. Because there are so many of them, I refer to them individually in the code otherwise I must always navigate across the full spec to find specific items. Conformance classes are not verbose enough in some cases about which specific requirement a part of the code implements or validates.

jerstlouis:
@fmigneault The requirement to execute a process is `/req/core/process-execute-op` in the published 1.0 version that you linked to. `/req/core/job-creation-op` was only used in the Abstract Test and is a broken link because it was not updated.

My PR does change one requirement label from `/req/core/process` to `/req/core/process-description` (or should it be /req/core/process-description-op`?) (Requirement 13 and maybe 14 as well?) to match the requirement file name, but we can discuss whether that change should be made or not.

fmigneault:
Indeed, you are right. Good to go for the change then since links are broken anyway.

Yes, maybe it is better to align the `...-op` name to keep them consistent.

pvretano (approved):

PR #288, opened 2022-04-29
fix required 'value' for 'measurement' in example process description

fmigneault:
Required field did not make sense for an input as :

```json
{ 
  "inputs": {
    "measureInput": {
      "value": {
        "measurement": 123,
        "uom": "m",
      }
    }
  }
}


pvretano:
Oopsie ... merging.

PR #289, opened 2022-06-09
Reactivates the auto build action workflow

ghobona:
Re-activated at the request of @pvretano

Cc: @ronaldtse @gbuehler


ronaldtse:
@ghobona I've just checked that this generate.yml is up-to-date with best practices. Thanks!

PR #290, opened 2022-06-14
Add support for URL-encoded execute requests (Issue 96)

pvretano:
Just the initial checking.  Need more review and changes.


christophenoel:
I agree that the distinction should be made clear in the specfication.

However, I would prefer style=form and explode=false instead of deepObject (which by the way is stated as only for objects, with no mapping for primitve value or array).

PR #291, opened 2022-07-21
Async execution response: 201 rather than 200

jerstlouis:
Shouldn't the async execution response code be 201 rather than 200?
(as it is currently in [7.11.4 Table 11](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc32))

### DIFF
--- a/core/sections/clause_7_core.adoc
+++ b/core/sections/clause_7_core.adoc
@@ -1039,7 +1039,7 @@ The following table maps the possible responses based on the combinations of the
 |async
 |_n/a_
 |_n/a_
-|200
+|201
 |_Location_ to newly created job
 |application/json
 |<<schema-statusInfo,statusInfo.yaml>>


fmigneault:
I believe it could be either 201 to indicate the creation of the job, or 202 to indicate the job was accepted but not yet processed (eg: it is pending execution). Either way, those would be more specific than 200.

pvretano:
@jerstlouis I agree with @fmigneault.  Can you update the PR to indicate either 201 or 202?

jerstlouis:
@fmigneault @pvretano I think 1.0 specifically requires 201 ([Requirement 34](https://docs.ogc.org/is/18-062r2/18-062r2.html#toc32)) (also Table 11)

To minimize changes I would suggest to stick with this, unless we really want to add that flexibility?

christophenoel:
> I think the argument for 201 rather than 202 was that the job resource is in fact already created, even if the results are still processing.
I still agree with 201 :)

PR #294, opened 2022-09-28
Clarify that a job dismiss should not touch results that perhaps have…

kalxas:
… been staged-out to external components

There has been a discussion in the EOEPCA project about whether to delete (or not) the processing results of a process when a dismiss operation is called. It feels natural that the files that have been stored on the processing server should be deleted on dismiss, but in the case that the processing results have been stored on an external storage (e.g. an s3 bucket) and perhaps have been registered to an external catalogue, the dismiss call should not delete those assets by default.
I am proposing this minor change to make this a bit more clear.

### DIFF
--- a/core/sections/clause_14_dismiss.adoc
+++ b/core/sections/clause_14_dismiss.adoc
@@ -2,7 +2,7 @@
 [[Dismiss]]
 == Requirements Class "Dismiss"
 
-The Dismiss requirement class specifies how to dismiss a job. Dismiss can be seen as cancelling a running job or removing artifacts of a finished job.
+The Dismiss requirement class specifies how to dismiss a job. Dismiss can be seen as cancelling a running job or removing internal artifacts of a finished job.
 
 include::../requirements/requirements_class_dismiss.adoc[]
 


bpross-52n:
The SWG discussed this on December 12th. We think that we need to split up the dismiss conformance class. We think about having one opertaion to cancel a running job and another one to delete the results of a finished job. Please leave us a comment about what you are thinking.

kalxas:
Thank you for the feedback.
Even if we have a "cancel" and a "dismiss" function separately, my above comment still stands: should externally stored results be deleted by the server?

gfenoy:
I made the two operations a single one as it was done in the [Dismiss extention](https://docs.opengeospatial.org/is/14-065/14-065.html#86) from WPS 2.0.

rconway:
The distinction between 'cancel' and 'delete' may be useful...

- CANCEL - stop the job execution, but leave the job in the list (with status cancelled or similar)
- DELETE - stop (if necessary) the job execution and remove the job (metadata) from the job list. iaw. WPS 2.0 dismiss, the server **may** also delete the assets (results) associated to the job

fmigneault:
CRIM's implementation also performs both cancel/dismiss tasks at the same time because there is no reason to leave results once the job is cancelled. In our case, even if the job was completed (successfully, cancelled or failed), a following dismiss request would NOT delete it, only change its status to `dismissed` and making sure to wipe result artifacts. The reason we preserve the job is to provide better errors messages such as HTTP Gone (410) instead of Not Found (404) when their status or results are requested afterwards.

bpross-52n:
@pvretano Will work on splitting the dismiss conformance class.

pvretano:
21-AUG-2023:  Discussed this in the SWG and the sentiment is that this conformance class should allow a job to be cancelled or deleted.  However, the conformance class should not say anything about what an implementation does with the the resources (internal and external) created by the job.  There should be a statement that basically says that it is up to the implementation to determine what to do with the artifacts created by the job.  Some implementations (e.g. CubeWerx or Zoo) delete all resources associated with the job.  Others (e.g. Hexagon) keep them around depending on the process.  So, it seems the behaviour is implementation dependent.  @pvretano will modify the PR to make the changes discussed.

bpross-52n:
> 21-AUG-2023: Discussed this in the SWG and the sentiment is that this conformance class should allow a job to be cancelled or deleted. However, the conformance class should not say anything about what an implementation does with the the resources (internal and external) created by the job. There should be a statement that basically says that it is up to the implementation to determine what to do with the artifacts created by the job. Some implementations (e.g. CubeWerx or Zoo) delete all resources associated with the job. Others (e.g. Hexagon) keep them around depending on the process. So, it seems the behaviour is implementation dependent. @pvretano will modify the PR to make the changes discussed.

SWG meeting from Dec 11th: make it so

pvretano:
Resolved in #398

PR #296, opened 2022-10-04
Proposal OGC API - Processes - Part 2 updates

gfenoy:
When we were testing the current OGC API - Processes - Part 2: Deploy, Replace, Undeploy schemas we faced various issues.

Some `$ref` were defined using the previous name of the GitHub repo (wps-rest-binding). 
The [ogcapppkg.yaml](https://github.com/opengeospatial/ogcapi-processes/compare/master...GeoLabs:ogcapi-processes:proposal-dru-updates?expand=1#diff-1da5d8e84254d7c4b793624365c113ebe7e2eb8bc55531a289fe8382a4a965b2) schema was not usable without modification.
The added [staticIndicator.yaml](https://github.com/opengeospatial/ogcapi-processes/compare/master...GeoLabs:ogcapi-processes:proposal-dru-updates?expand=1#diff-86aaa8f65d8695c32bcfb6d17329c5eca80f21a29c3ccbbd8123d8e030d2ad65) schema extends the [processSummary.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/core/openapi/schemas/processSummary.yaml) one as expressed in [immutable processes section](https://github.com/opengeospatial/ogcapi-processes/blob/master/extensions/deploy_replace_undeploy/standard/sections/clause_6_deploy_replace_undeploy.adoc#immutable-processes).

Also, a specific `swagger` folder has been added for being able to use the schema from swagger-ui as illustrated [here](http://tb18.geolabs.fr:8080/swagger-ui/oapip/).

### DIFF
--- a/extensions/deploy_replace_undeploy/standard/openapi/responses/DeployProcess.yaml
+++ b/extensions/deploy_replace_undeploy/standard/openapi/responses/DeployProcess.yaml
@@ -1,0 +1,13 @@
+description: the process is deployed
+headers:
+  Location:
+    schema:
+      type: string
+    description: URL to fetch the processDescription of the deployed process
+content:
+  application/json:
+    schema:
+      content:
+        application/json:
+      schema:
+        $ref: "../schemas/staticIndicator.yaml"
--- a/extensions/deploy_replace_undeploy/standard/openapi/responses/DuplicateProcess.yaml
+++ b/extensions/deploy_replace_undeploy/standard/openapi/responses/DuplicateProcess.yaml
@@ -1,8 +1,8 @@
 description: the processes being added is already deployed (i.e. duplicate)
 content:
-  application/json:
+  application/problem+json:
     schema:
-      $ref: "https://raw.githubusercontent.com/opengeospatial/wps-rest-binding/master/core/openapi/schemas/exception.yaml"
+      $ref: "../../../../../core/openapi/schemas/exception.yaml"
   text/html:
     schema:
       type: string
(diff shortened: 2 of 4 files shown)


fmigneault:
The modified `$ref` are very error-prone with all the relative `../../..` references. They should probably use the full URI instead `https://schemas.opengis.net/ogcapi/processes/part1/1.0/openapi/...`

fmigneault:
Looks good now.

fmigneault (changes requested):
I revoke my previous approval due to changes introduced in https://github.com/opengeospatial/ogcapi-processes/commit/12f5192181906296e368ebe33bf7e100a8837b18. 
The change from `executionUnit` array to object will completely break existing instance deployments. This is a major backward incompatibility that cannot be integrated as `1.x` version.

gfenoy:
> I revoke my previous approval due to changes introduced in [12f5192](https://github.com/opengeospatial/ogcapi-processes/commit/12f5192181906296e368ebe33bf7e100a8837b18). The change from `executionUnit` array to object will completely break existing instance deployments. This is a major backward incompatibility that cannot be integrated as `1.x` version.

Thanks for reporting the issue @fmigneault. In our implementation, we also support an array for ``executionUnit``. 

We made this modification due to a remark made during a meeting about Part 2 and Part 3.

I am in favor of going back to the initial array support.

fmigneault:
@gfenoy
I don't mind if both are permitted, as long as the previous array representation remains available. 
I think it would be worth it to also include explicitly the `type`/`href` vs `mediaType`/`value` fields, since they are essential to properly parse the contents.
Previous testbeds were using `href`/`unit` instead of `href`/`value`, but I like the alignment of `value` with other places that use those same names, like in metadata references.

PR #318, opened 2023-02-13
Patch missing label issues.

pvretano:
I've patched the files mentioned in this issue but the use of the metanorma "label", "part", etc. directives in the requirements and reccomendations is quite inconsistent.


jerstlouis:
@pvretano You may already know this, but just in case, I believe the `label` is the requirement ID (and was more recently replaced by `identifier`) e.g., for _Requirement 1_: `/req/core/landingpage-op` whereas `part` is to identify multiple parts e.g., A, B within _Requirement 10_.

> The identifier was previously encoded as label until https://github.com/metanorma/metanorma-ogc/releases/tag/v2.2.0

from https://www.metanorma.org/author/topics/document-format/requirements-modspec/

pvretano:
@jerstlouis nope, I was not aware of that.  However, as I mentioned about after the metanorma migration that tags used to label identifiers, parts, etc. and how exactly they are encoded in a file are all over the place.  Since I don't have time to read the metanorma documentation ... nor do I care to do so ... someone else more familiar with that product will have to do a sweep of all the recommendations and requirements and harmonize eveything.  This is something that should be done closer to the time we anticipate handing the document over to the OAB for review.

PR #324, opened 2023-05-01
BBOX, Example server #1, README

jerstlouis:


jerstlouis:
Merged.

fmigneault:
Just curious. 
What's the point of the `anyOf` if the enum is short-circuited with the generic string URI format?
Can't the definition be simplified only to that URI format, and move the enum values to `examples`?

pvretano:
@fmigneault normally that enum list would be augmented with the list of CRSs that the server actually supports.  However, should anyone extent the list and they try validation using the schemas at opengis.net, validation will fail.  The short circuit is to allow validation again the normative schemas at opengis.net to succeed when the server supports more than just CRS84 and CRS84h.  @jerstlouis can probably explain it better than me.

jerstlouis:
@fmigneault We discussed this in the call this morning. There is a permission that says that implementations can add additional values to the "enum" to declare the CRSs that they support from the OpenAPI definition.
At the moment, there is no other way to do so.

I added an additional note that clarifies that the reason for the `anyOf` is so that execution requests will validate against the OGC canonical schema.

Issue #319, opened 2023-02-24
Inconsistent schemas in DRU extension since OpenAPI restructure


fmigneault:
Since #306, both of these files exist at the same time:
- https://github.com/opengeospatial/ogcapi-processes/blob/master/extensions/deploy_replace_undeploy/standard/openapi/schemas/ogcapppkg.yaml
- https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-dru/ogcapppkg.yaml

Although their names are the same, their contents are entirely different.

Multiple other items under https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/deploy_replace_undeploy/standard/openapi have not been moved at all. 

Having both locations with inconsistent schemas makes it very confusing for referencing this extension.

jerstlouis:
@fmigneault @gfenoy 

When I worked on #306, I updated all of the reference from Processes - Part 1, but not those from Part 2.
That is why I left the schemas in Part 2 until it is updated properly.
I am not sure what the ogcapppkg in the root openapi directory is -- it might be the updated one from Gerald.
Gerald could you please verify and do a PR to update Part 2 to reference the new one, and remove the old schemas once everything works fine?

Note that the top-level metanorma document for Part 2 will have to be moved to the root alongside Part 1 & 3 to avoid errors referencing above it.

gfenoy:
@fmigneault thanks for your patience.

The schemas should now be correct in the root openapi location [62f3743](https://github.com/opengeospatial/ogcapi-processes/pull/327/commits/62f374335fd5deb5838052107afb6123cb1959ef). Also, I have modified the [ogcapi-processes.yaml](https://github.com/GeoLabs/ogcapi-processes/blob/feature/fix-issue-319/openapi/ogcapi-processes.yaml) to include Part 2, but did not push the generated json bundle yet.

fmigneault:
@gfenoy 
No problem. Thanks for working on this!

Would it be possible to include the `application/cwl`, `application/cwl+json` and `application/ogcapppkg+json` Content-Types for Deploy body as well to support the use-cases defined in these references?
- https://github.com/opengeospatial/ogcapi-processes/blob/master/extensions/deploy_replace_undeploy/standard/recommendations/deploy-replace-undeploy/deploy/REC_body-cwl.adoc
[...]

PR #327, opened 2023-05-30
Fix proposal for #319

gfenoy:
This changes should solve the issue #319 related to moving the schemas location from the extension specific openapi subdirectory to root.


jerstlouis:
Thanks @gfenoy !

We should also add the paths for _processes-dru_ to [openapi/paths](https://github.com/opengeospatial/ogcapi-processes/tree/master/openapi/paths) and then include them in [ogcapi-processes.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/ogcapi-processes.yaml
). Then if it works with your demo API, we could add your server as an example supporting DRU to the [list of example servers](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/ogcapi-processes.yaml#L13).

You could also potentially use the example OpenAPI schemas for your implementation to assemble your API definition, with `swagger-cli bundle ogcapi-processes.yaml`. Simply comment out any paths that your API doesn't implement, tweak the [process-list.json](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/api/processes-list.json)  to the processes your API supports (or return it dynamically from your server at`/api/process-list`), and add any extra things that you support as custom extensions or from other OGC APIs (several of which now following this structure).

openapi/ogcapi-processes.yaml
pvretano, line 1: Why `pDeploy.yaml` and `pDelete.yaml`?  The naming is a little odd!  I would just name the files `Deploy.yaml` and `Delete.yaml`.

gfenoy: You are perfectly right, this is why I have renamed pDelete.yaml to pUndeploy.yaml.

For the `p` prefix, I thought it was the new way of defining paths with the new location, comes new naming also:
* https://github.com/opengeospatial/ogcapi-processes/tree/master/openapi/paths/processes-core
* https://github.com/opengeospatial/ogcapi-processes/tree/master/openapi/paths/processes-workflows

So, I simply followed the same naming scheme.

jerstlouis: @gfenoy @pvretano The _p_ prefix is for _paths_, the _r_ prefix is for _responses_.
It is to distinguish the files between them and from the `_schemas_`  that are not prefixed,  as you the same name may often be used in both directories and it avoids the confusion when only the filename is seen.

gfenoy: I have to rename the responses.

The names for the initial schemas were named already, so I only move them.

I will fix that right now.

gfenoy:
I have made the renaming using the `r` prefix for responses

openapi/ogcapi-processes.yaml
jerstlouis: They are already in the respective directories.

The purpose is to distinguish between e.g.,

https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/responses/common-core/rLandingPage.yaml

and

https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/paths/common-core/pLandingPage.yaml

The important thing is to have different filenames to avoid confusion.
Users not familiar with the standard may never see these. Only the schemas are referenced from the Standard.
It all gets bundled up in a single JSON API definition file in the end.
They only need to edit the main ogcapi-processes.yaml.

I found this naming convention to be extremely useful in avoiding confusion when putting all of this together in all the different specs where this approach is used.

jerstlouis:
@gfenoy Now you can add your server to the list of examples and say it supports DRU :) Then we can test it all directly from  SwaggerUI.

gfenoy:
@jerstlouis should I also push the bundle to replace the current one?

I ask this question because I noticed that there is no mention of Part 3 in the ogcapi-processes.

Also, we may think of having the following files:

* openapi/ogcapi-processes-1.yaml: include the current ogcapi-processes.yaml file (before the modifications I made)
* openapi/ogcapi-processes-2.yaml: include the ogcapi-processes.yaml offering both part 1 and part 2 definitions
* openapi/ogcapi-processes-3.yaml: include the ogcapi-processes.yaml offering both part 1, part 2 and part 3 definitions

What do you think?

openapi/ogcapi-processes.yaml
pvretano: Put them in a sub directory under the sub directory. I don't think the p and r prefixes are clear. otherwise rename the files to be something like DeployPath.yaml or UndeployResponse.yaml ... although I think paths and responses sub directories are better.

jerstlouis:
@gfenoy Yes you can update the updated bundle.

> t there is no mention of Part 3 in the [ogcapi-processes.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/ogcapi-processes.yaml).

That's because Part 3 does not realy define any new paths, it only extends the execution request and defines the `response=collection/landingPage` on the POST to `/execution`.

> Also, we may think of having the following files:

I think it's simpler to just include everything, and developers can comment out the paths they do not implement.
This is why I suggest it would be great if you can add your server which implements Part 2 as an additional API deployment example, since our server that is currently the only one there in the list only supports Part 1 and 3.

openapi/ogcapi-processes.yaml
pvretano: Sorry @jerstlouis but I find it very confusing.  The very path you you provided has the word `responses` in it and to the `r` is completely meaningless and superfluous.  I would be happy to create a PR to rename the files.  If you want to know that a file independent of its path (does that happen often?) is related to responses or paths and use full words in the name.  rDeploy.yaml means nothing!  DeployResponse.yaml is clear but in my opinion unnecessary since the path to the file already identifies what it does.

The "because Tiles does it this way" agument is a bit weak because there are plenty of other OGC specifications that don't use this convention so the topic of consistency on this point is moot.

gfenoy: > This is why I suggest it would be great if you can add your server which implements Part 2 as an additional API deployment example

I will proceed after we get confirmation that once deployed a Server Instance relying on the proposed new schemas is properly working.

gfenoy:
> This is why I suggest it would be great if you can add your server which implements Part 2 as an additional API deployment example

I will proceed after we get confirmation that once deployed a Server Instance relying on the proposed new schemas is properly working.

openapi/ogcapi-processes.yaml
jerstlouis: @pvretano at this point it's already OGC API - Coverages, Tiles, Maps, DGGS and Processes, so that is five of them.

Having these consistent makes two important things much easier:
- Synchronizing the common building blocks between them (and with _Common_) by simply doing a directory comparison with e.g., Meld
- Assembling an OGC API definition that supports those different OGC API for your implementation by combining these directories that fit seamlessly together

_rDeploy.yaml_ means something once you've understood that _r_ is the prefix for responses.

These paths and responses files are not something highly visible to the implementers and readers of standards, but they are to the standards editor.
I don't really want to argue strongly for one way or the other, but the consistency points above I think are very important.

pvretano: Don't add Processes to the list just yet!  Clearly you and I don't agree so I think we need to let the rest of the SWG weigh in with a SWG vote on the matter.
Also, the new scheme you proposed does indeed help with synchronization and building OpenAPI document across multiple APIs.  I don't tjhink `p` or `r` helps or hinders that and if we must make that distinction then adding `Path` and `Response` suffixies is much, much clearer.
If you are OK with a SWG vote, I'll send @bpross-52n and email requesting he set up an eVote.

jerstlouis: I do have a preference for the _p_ and _r_ preffxes, but I really wouldn't mind that much (even if it meant I had to change all of the other aforementioned 4 other standards) if not for breaking the pattern from the already published _OGC API - Tiles_.

I don't really see why you hate those prefixes so much as to call for an eVote on the matter, but OK :)

pvretano: @jerstlouis I don't "hate" it but I also don't see why you include the prefixes when in every case I can think of the files are referenced using their path and the path already contains that information.  I was not proposing you change the other specifications at all ... just Processes.  However, if it means changing an already published standard (not simple) then just leave it.

openapi/responses/processes-dru/rDeployProcess.yaml
pvretano, line 1: What does the `r` mean again! ;)

jerstlouis: These prefixes are grrrrrreat ;) (-Tony the Tiger)

pvretano: LOL!

openapi/README.md
```diff
@@ -8,7 +8,7 @@ The list of supported paths should be ajusted in `ogcapi-processes-3.yaml`.
 
 The `ogcapi-processes-3.bundled.json` was generated with `swagger-cli bundle` from `ogcapi-processes-3.yaml` and its dependencies included from the components sub-directories.
 
-The `paths/processes-dru/pProcessListDeploy.yaml` and `paths/processes-dru/pProcessDescriptionReplaceUndeploy.yaml` files are generated by concatening multiple files in a single one. you can sue the command bellow to update their value with the current schema:
+The `paths/processes-dru/pProcessListDeploy.yaml` and `paths/processes-dru/pProcessDescriptionReplaceUndeploy.yaml` files are generated by concatening multiple files in a single one. you can use the command bellow to update their value with the current schema:
```
jerstlouis, line 11: Thanks @gfenoy . _bellow_ is still wrong though :)

gfenoy: Thanks for you patience @jerstlouis

jerstlouis: @gfenoy No thank you for yours ;) Sorry for being so annoying finding things to fix :P

openapi/parameters/processes-dru/w-param.yaml
```diff
@@ -0,0 +1,5 @@
+name: w
```
jerstlouis, line 1: @gfenoy What's `w` for? Can we add a `description`?

jerstlouis: Thanks.

fmigneault:
@gfenoy 
See https://github.com/opengeospatial/ogcapi-processes/pull/329 for the CWL schema discussed in #319.

gfenoy:
@fmigneault following the discussion [there](https://github.com/opengeospatial/ogcapi-processes/pull/329#issuecomment-1580394503).

My understanding is that the schema will be made available from there https://github.com/common-workflow-language/cwl-v1.2/.

I don't know yet if we we plan to keep a copy of the schema also in this repository. I would like to have a confirmation about this.

If not, then I will wait for the schema to be made available there to update the schema to reference this schema.

fmigneault:
@gfenoy 
I think it would be better to use the published external reference to make sure it stays up-to-date with any fixes applied on the CWL side.

fmigneault:
@gfenoy @jerstlouis 
The official CWL JSON schema is now available here: 
https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/json-schema/cwl.yaml

openapi/paths/processes-dru/pDeploy.yaml
```diff
@@ -0,0 +1,34 @@
...
+     content:
+       application/ogcapppkg+json:
+         schema:
+           $ref: "../../schemas/processes-dru/ogcapppkg.yaml"
+       application/cwl:
+         schema:
+           type: string
+       application/cwl+json:
+         schema:
+           example: {}
```
fmigneault, line 25:
```suggestion
       application/cwl:
         schema:
           $ref: "https://raw.githubusercontent.com/common-workflow-language/cwl-v1.2/1.2.1_proposed/json-schema/cwl.yaml"
       application/cwl+json:
         schema:
           $ref: "https://raw.githubusercontent.com/common-workflow-language/cwl-v1.2/1.2.1_proposed/json-schema/cwl.yaml"
       application/cwl+yaml:
         schema:
           $ref: "https://raw.githubusercontent.com/common-workflow-language/cwl-v1.2/1.2.1_proposed/json-schema/cwl.yaml"
```

fmigneault:
@gfenoy 

In these two definitions, I recall having a nested `unit` or `href` mapping that allowed respectively to either pass the definitions direct;y as JSON, or provide an external reference to a document containing the execution unit contents.

https://github.com/GeoLabs/ogcapi-processes/blob/a8347b5f2c9e37e4581b68739120e62c4ea1317a/openapi/schemas/processes-dru/ogcapppkg-array.yaml#L3-L6

https://github.com/GeoLabs/ogcapi-processes/blob/a8347b5f2c9e37e4581b68739120e62c4ea1317a/openapi/schemas/processes-dru/ogcapppkg.yaml#L7-L12

Also, https://github.com/common-workflow-language/cwl-v1.2/blob/1.2.1_proposed/json-schema/cwl.yaml can now be added to those lists as well.

fmigneault:
@gfenoy Is there something still blocking this PR?

gfenoy:
@fmigneault we were discussing about merging this PR during the last SWG meeting and no objection was raised.

I plan to attend the today meeting and hope we can get this PR merged during the SWG meeting except if an objection is exposed.

In case you have time, it would be great to have you attending the meeting to support this PR.

pvretano:
21-AUG-2023: SWG reviewed the PR and agreed to merge.  If any issues arrive after the fact, new issues can be created to address those.

PR #333, opened 2023-07-13
Fix successUrl -> successUri.

guidorice:
This PR fixes the naming of a required property on the subscriber schema.


jerstlouis:
@pvretano Is this only a typo fix, or could this potentially be a breaking change?

The OGC API - Processes - Part 1: Core standard also seems very light on actual requirement for this callback functionality.

It references https://swagger.io/docs/specification/callbacks/ which does include a `successUrl` with an L (URL).

gfenoy:
I think that the issue comes from the current schema located here: https://schemas.opengis.net/ogcapi/processes/part1/1.0/openapi/schemas/subscriber.yaml

It mentions that the required value is one which is not available `successUrl` when `successUri` is defined as one possible option.

PR #360, opened 2023-09-18
add CRIM Weaver server/client components

fmigneault:


fmigneault:
@bpross-52n as requested during the kickoff

PR #378, opened 2023-11-13
Initial checkin to resolve issue 373.

### LINKED ISSUE #373: Ambiguity around array inputs?
If I want to provide an array of max. two strings, how am I supposed to do that?

1) Purely JSON Schema:
```yaml
inputs:
  example:
    schema:
      type: array
      maxItems: 2
      items:
        type: string
```

2) maxOccurs:
```yaml
inputs:
  example:
    maxOccurs: 2
    schema:
      type: string
```

Is it equivalent?

3) And what does the following mean? 
```yaml
inputs:
  example:
    maxOccurs: 2
    schema:
      type: array
      maxItems: 2
      items:
        type: string
```

(Provided in YAML for simplicity)

pvretano:
Closes #373


core/examples/json/ProcessDescription.json
```diff
@@ -151,21 +150,24 @@
...
-          {
-            "type": "string",
-            "contentEncoding": "binary",
-            "contentMediaType": "image/jp2"
-          }
-        ]
+        "type": "array",
+	"minItems": 1,
+	"maxItems": 150,
+        "items": {
```
fmigneault, line 157: Missing the single item (non-array) representation.

core/sections/clause_6a_deprecated.adoc
```diff
@@ -0,0 +1,29 @@
...
+[source,YAML]
+----
+schema:
+  oneOf:
+    - <original-schema>
+    - type: array
+      items: <original-schema>
+      minItems: 1
+      maxItems: N
+----
```
fmigneault, line 27: One issue with this is that it will not be possible to distinguish whether this `schema` represents the previous 1.0 `minOccurs=1`/`maxOccurs=1` (which could be omitted), or the new (2.0?) strategy to represent cardinality.

Issue #373 doesn't seem resolved with this approach alone, but rather made more ambiguous, if one intends to be backward compatible. 

A process description using 2.0 might need to always report `minOccurs=1` and `maxOccurs=1` explicitly to disambiguate.

core/sections/clause_7_core.adoc
```diff
@@ -504,12 +504,151 @@ NOTE: This schema can also be obtained from https://raw.githubusercontent.com/op
...
+.Cardinality rules for process inputs.
+[cols="10,10,60,40",options="header"]
+|===
+|minOccurs |maxOccurs |Interpretation |Example
+|0 |0 |Not Allowed |
+|0 |1 |A single input value *may* be optionally specified. | "input": {value}
+|1 |1 |A single input value *must* be specified. | "input": {value}
+|1 |N |At least 1 input value *must* be specified.  +
+Regardless of the number of input values provided, those values *must* be encoded using an array. | "input": [{value}] OR +
+"input": [{value1},...,{valueN}]
```
fmigneault, line 538: This is not backward compatible (the value cannot be provided directly without array nesting). Is it intentional?
If so, my previous comment about "missing single item (non-array) representation" can be discarded.

core/sections/clause_7_core.adoc
```diff
@@ -504,12 +504,151 @@ NOTE: This schema can also be obtained from https://raw.githubusercontent.com/op
...
+  input:
+    schema:
+      type: array
+      maxItems: 2
+      items:
+        type: string
+----
 
-The name of each input is its identifier as specified by the input's definition in the <<sc_process_description,process description>>.
+In this case the schema on an input value is defined as an array and so will always be encoded as an array in an execute request.
```
fmigneault, line 568: Should cardinality be considered of 1 or 2 here?

If cardinality is 1 (ie: a single input that happens to be a value represented as an array of max 2 items), then how does one differentiate this array from the `imagesInput` example that uses the same representation for cardinality >1.

If cardinality is 2 (ie: 1 or 2 distinct values passed as an array), how should one represent a single "complex" input that is represented as an array allowing at most 2 sub-elements?

core/sections/clause_7_core.adoc
```diff
@@ -504,12 +504,151 @@ NOTE: This schema can also be obtained from https://raw.githubusercontent.com/op
...
+"input": [{value1},...,{valueN}]
+|M |N |All values provided *must* be encoded using an array. |"input": [{value1},...,{valueM}] OR +
+"input": [{value1},...,{valueM},...,{valueN}]
+|===
+
+NOTE: The tokens `{value}`, `{value1}`, `{valueN}` or `{valueM}` represent values of the type expected for the input according to its definition.  This can include array values.
+
+====== Interaction of `minOccurs`/`maxOccurs` and `minItems`/`maxItems`
+
+The `schema` member of the <<process-input-schema,input definition>> in the <<sc_process_description,process description>> defines the schema of a single instance of an input.  If the input happens to be an array then the `minItems` and `maxItems` properties may be used to define the limits of the array.  The use of `minItems` and `maxItems` in the definition of the input does not affect how `minOccurs` and `maxOccurs` are interpreted and the same <<cardinality_rules,cardinality rules>> apply.
```
fmigneault, line 547: This section seems to describe how cardinality is represented currently (1.0).

This doesn't align however with the `schema` using only `minItems` and `maxItems` in the `imagesInput` example to indicate the full cardinality representation, not a single instance of it. 

Doesn't this cause even more ambiguity regarding #373?

fmigneault (commented):
The "Cardinality" chapter seems very appropriate to explain the current 1.0 behavior, but it contradicts with the updated examples that use the planned deprecation of `maxOccurs`.

core/examples/json/ProcessDescription.json
pvretano: I'm updating this PR to undo the deprecation of minOccurs and maxOccurs.

core/sections/clause_6a_deprecated.adoc
pvretano: I am updating this PR to undo the deprecation of minOccurs and maxOccurs.

core/sections/clause_7_core.adoc
pvretano: @fmigneault I'm trying to find out how this is not  backward compatible.  I see Requirement 19 in the current, published, version which seems to say the same thing as lines 536 to 538 ... no?

core/sections/clause_7_core.adoc
pvretano: Yeah, we discussed in the SWG and decided not to deprecate minOccurs and maxOccurs since this action, as you say, causes more ambiguity.  As a result, I am updating this PR to clarify minOccurs and maxOccurs but not to deprecate them.

core/sections/clause_7_core.adoc
fmigneault: It was only problematic to disambiguate between previous (v1) and the original proposition to remove min/max-occurs assuming the default to omit them would not align with explicitly omitting them since they would not be supported. 
Since the decision is to keep them, then this is not an issue anymore.

m-mohr:
Is there a way to translate minOccurs/maxOccurs into a JSON Schema equivalent without them? If there is, it might help to have a note how to translate between them somewhere in the spec.

core/sections/clause_7_core.adoc
pvretano: Done ... I hope! ;)

core/sections/clause_7_core.adoc
pvretano: The images input example has been updated to use minOccurs/maxOccurs.

core/examples/json/ProcessDescription.json
pvretano: The `imagesInput` example now uses minOccurs/maxOccurs.

core/sections/clause_6a_deprecated.adoc
pvretano: clause_6a_deprecated.adoc has been removed.

core/sections/clause_7_core.adoc
pvretano: Closing as per @fmigneault last comment.

PR #379, opened 2023-11-20
fix invalid reference adoc OGC AppPkg schema

fmigneault:
The `ogcapppkg` schema reference seemed to be missing one `../` to find the correct location.


fmigneault:
@pvretano FYI

PR #385, opened 2023-12-11
Value passing in an execute request.

### LINKED ISSUE #369: Execution endpoint input specification should be separate conformance class?
I know our implementation is probably designed a bit differently than a "normal" OGC Processes implementation, but we've run into a bit of a hiccup getting conformance with core due to the complexity specifying inputs when executing a workflow.

For us, we are currently passing a single input into all our processes that is always a json object, more specifically a STAC item collection. We currently have no use for and do not currently want to support things like passing inputs by reference, as we effectively always get the inputs by reference because they are STAC (well, they're not always valid STAC (like when a STAC item hasn't ever existed and we are trying to make one), but we say the inputs will always be "STAC-like" and require the use of references to an underlying data assets like STAC). And we don't need to have single value parameters either--anything like that is also embedded in our json payload format.

So we end up passing this input as a qualified input value in all cases, which works of course. But it means that we only need to support that one branch of all the several ways the spec supports specifying inputs, and we don't want to add code we won't use and don't want to support just to get full conformance with the spec.

This latter point suggests to me that perhaps this input specification part of the core conformance class is not actually "core"--and thus I wanted to see if there might be interest in considering the idea of splitting it out into a separate conformance class or set of classes. Maybe there'd be a different conformance class for input values vs references vs qualified inputs?

pvretano:
Allow value passing to be specified per input rather than requiring thes erver to support both by-value and by-refernce in all cases.  The default value for the `valuePassing` parameter is `["byValue","byReference"]` to maintain backward compatability.

Closes #369

### DIFF
--- a/core/requirements/core/REQ_process-execute-inputs.adoc
+++ b/core/requirements/core/REQ_process-execute-inputs.adoc
@@ -5,12 +5,22 @@
 identifier:: /req/core/process-execute-inputs
 [.component,class=part]
 --
-The server SHALL support process input values specified in-line in an execute request (i.e. by value).
+How process input values can be specified in an execute request (by-value, by-reference or both) SHALL be controlled by the https://raw.githubusercontent.com/opengeospatial/ogcapi-processes/master/openapi/schemas/processes-core/inputDescription.yaml[`valuePassing`] parameter that may be specified per input in the <<sc_process_description,process description>>.
+--
+
+[.component,class=part]
+--
+The default value of the https://raw.githubusercontent.com/opengeospatial/ogcapi-processes/master/openapi/schemas/processes-core/inputDescription.yaml[`valuePassing`] parameter shall be `["byValue","byReference"]`.
+--
+
+[.component,class=part]
+--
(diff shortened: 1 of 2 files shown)


fmigneault:
@pvretano 
Sorry I couldn't assist this morning. Can you add precision about `valuePassing`?
Is it an extra parameter to be added for each input of each process description?
Can the server specify a conformance class it supports server-wide in the `conformsTo` response?

jerstlouis:
@fmigneault My understanding is that it is to be explicitly specified for **each input of each procses** description

pvretano:
@fmigneault, @jerstlouis is correct.  The `valuePassing` parameter can be specified for each input in a process description.  If it is not specified the default value defined by the specification is `["byValue","byReference"]` which is backward compatible with the current version of processes.  

Question: Would it be of value to also add a `valuePassing` parameter at the top level of the process description to be a sort of global default?  It would alleviate having to repeat the same values over and over and to also override the specification default.

jerstlouis:
@pvretano My opinion is that would not be very helpful, because the supported values are very likely to depend a lot more on the type of inputs.

e.g., a number or text input would normally be supported "byValue" and a GeoTIFF "byReference".

fmigneault:
I agree. I prefer to have them repeated in each input than having to cross-validate with the process description each time.

bpross-52n:
SWG Meeting from January 22nd: Remove min/max items and make the enum extensible.

PR #388, opened 2024-01-05
Fature/apply review fixes

gfenoy:
Modifications were required after the review was sent to the SWG. This PR should solve most reported issues and clarifications.

Changed from recommendation to Requirement, updated original recommendation:
*  /rec/ogcapppkg/execution-unit-cwl -> /reg/cwl/execution-unit
*  /rec/ogcapppkg/execution-unit-docker -> /reg/docker/execution-unit

Split recommendation into Requirement and recommendation:
* /rec/deploy-replace-undeploy/deploy/body-ogcapppkg -> /req/ogcapppkg/deploy/body
* /rec/deploy-replace-undeploy/deploy/body-cwl -> /req/cwl/deploy/body
* /rec/deploy-replace-undeploy/replace/body-ogcapppkg -> /req/ogcapppkg/replace/body
* /rec/deploy-replace-undeploy/replace/body-cwl -> /req/cwl/replace/body

Update Requirement references from the ATS consequently.


extensions/deploy_replace_undeploy/standard/requirements/deploy-replace-undeploy/deploy/REQ_w-param.adoc
pvretano, line 1: This should be removed from the core since it seems to a be a CWL-specific thing.  The core should be nice and simple.  The POST body contains the definition of a single process to be deployed.  Also, the name `w`, defined from "workflow" I suspect is also problematic.  All workflow-related concepts should be restricted to Part 3.

extensions/deploy_replace_undeploy/standard/requirements/cwl/deploy/REQ_w-param.adoc
pvretano, line 1: Hmm ... it might be something about CWL that I don't understand but I think this should be remove from the standard altogether.  The POST body for a deploy operation should contain the definition of a single process and that is it.  Nice and simple.  This means that a CWL body should contain a single process definition (or "workflow" in CWL speak I guess).  

However, as I said, this might be an effect of something about CWL that I don't understand.  If that is the case then please exaplain.

extensions/deploy_replace_undeploy/standard/requirements/deploy-replace-undeploy/deploy/REQ_w-param.adoc
gfenoy: The Requirement has been moved from the DRU to the OGC Application Package Requirements class.

* /req/deploy-replace-undeploy/deploy/w-param -> /req/ogcapppkg/deploy/w-param

As mentioned [here](https://github.com/opengeospatial/ogcapi-processes/pull/386#discussion_r1433343826) about the Requirement /deploy/w-param for another Workflow Language:
> Might be relevant to refer to a common parameter that can be reused across Workflow languages regardless of their specific implementation.

I agree with this. It will ease the integration of more Workflow Languages in the future. 

So, the CWL Requirements class refers to the Requirement /req/ogcapppkg/deploy/w-param for the Deploy operation.

In the [OGC Application Package schema](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-dru/ogcapppkg.yaml), the `executionUnit` property is defined as a oneOf where [array](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-dru/ogcapppkg-array.yaml) can be an option.

extensions/deploy_replace_undeploy/standard/requirements/cwl/deploy/REQ_w-param.adoc
gfenoy: In CWL, we can identify a process as an instance of the workflow class. This was clarified in https://github.com/opengeospatial/ogcapi-processes/commit/3105532ed3f2d342e2236edb9335b82a6a1292ea.

A CWL file can contain multiple workflow definitions. The w parameter permits the client to select the one to deploy as a process from the Processes Part 2 implementation.

As illustrated in PR #386, it is a concept shared among Workflow Languages that can be used to deploy processes.

fmigneault: In CWL, the contents can be defined under `$graph` which contains a list of multiple CWL "applications". 
Each of those "applications" can themselves be sub-workflows or atomic command line tools that can be combined into a main workflow. The `w` parameter can help disambiguate which one of IDs is the targeted tool for deployment as OGC API Process.

gfenoy: As discussed during the SWG meeting on January 8th,  2024, we moved the requirements w-param and exception-workflow-not-found back to the CWL Requirements class. They are not referenced from anywhere else.

Issue #282, opened 2022-03-01
Update Part 2 with the consistent use of process "definition" and process "description"


pvretano:
Update part 2 to reflect that fact that you:
GET a process **description** from `/processes/{processId}' (same terminology is used in part 1)
and you POST a process **definition** to `/processes` to deploy a process.  In other words a process definition is an application package (i.e. description + execution unit + possibly other stuff).

Also, a new endpoint `/processes/{processId}/package` for get the process definition OR we use the same endpoint (i.e. `/processes/{processId}`) and use content negotiation to get the description versus the definition.  There is already a MIME type for an application package so that should be straight forward.

jerstlouis:
@pvretano At the moment, the OGC API Application Package is only **one** potential language/format for the process **definition**.

If we keep it that way, neither of those options (calling it `/package` or apppkg media type) seem appropriate.

I feel that the definition and description are different enough that they should be separate, especially given the amount of confusion that mixing those up generate.

pvretano:
@jerstlouis application pkg as defined in PART 2 is NOT MANDATORY.  If it suited to your purpose it is recommended.  Other language/formats are possible and if another one emerges (e.g. CWL or MOAW) we can add that as another recommendation.

jerstlouis:
@pvretano Yes exactly, that's what I meant.

cportele:
I am not able to follow the Processes discussion that is happening, but I saw the issue and think it is strange to POST a representation of a resource of type A (process definition) and when you GET the resource that has been created you get a representation of a different resource type B (process description). This doesn't sounds like HTTP to me. I am probably missing the necessary context, so feel free to ignore...

fmigneault:
I'm fine with any of:
- GET `/processes/{processId}/package` 
- GET `/processes/{processId}/executionUnit` 
- GET `/processes/{processId}` + `Accept: application/ogcapppkg+json`
- GET `/processes/{processId}?f=ogcapppkg`

I will redirect all of these to whichever gets decided. 
If possible, I prefer to have path components over headers/queries.

It is important to ensure `Accept`/`f` are **required** if  `/processes/{processId}` is used for POST/PUT.

fmigneault:
@cportele 
It is mostly to preserve backward compatibility with *Core*. Deployed process or builtin are all obtained through GET `/processes/{processId}`. 
Any extra information needed to define the process can be POSTed *as well*, but not mandatory. 
[...]

PR #404, opened 2024-04-26
Fix issue #282

gfenoy:
As discussed on April 15, 2024, during the SWG meeting, we propose the following additions.

Add /processes/{processId}/package path for accessing the formal description used to deploy a process.

Add specific recommendations and requirements for the OGC Application Package and CWL conformance classes.

### DIFF
--- a/extensions/deploy_replace_undeploy/standard/requirements/cwl/package/REQ_response-body.adoc
+++ b/extensions/deploy_replace_undeploy/standard/requirements/cwl/package/REQ_response-body.adoc
@@ -1,0 +1,10 @@
+[[req_cwl_package_response-body]]
+[requirement]
+====
+[%metadata]
+label:: /req/cwl/package/response-body
+part:: A response with HTTP status code `200` SHALL include a body that contains:
+ * the <<rc_cwl,CWL>> to use to deploy the process, in case the Content-Type used to deploy the process was `application/cwl`. 
+ * the <<rc_ogcapppkg,OGC Application Package>> to use to deploy the process, in case the Content-Type used to deploy the process was `application/ogcapppkg+json`.
+ 
+====
--- a/extensions/deploy_replace_undeploy/standard/requirements/deploy-replace-undeploy/package/REQ_get-op.adoc
+++ b/extensions/deploy_replace_undeploy/standard/requirements/deploy-replace-undeploy/package/REQ_get-op.adoc
@@ -1,0 +1,9 @@
+[[req_deploy-replace-undeploy_package_get-op]]
+[requirement]
+====
+[%metadata]
+label:: /req/deploy-replace-undeploy/package/get-op
+part:: For every dynamically added process (path: /processes/{processId}), the server SHALL support the HTTP GET operation at the path `/processes/{processId}/package`.
+part:: The parameter `processId` is each `id` property in the process collection response (JSONPath: `$.processes[*].id`).
+
+====
--- a/extensions/deploy_replace_undeploy/standard/requirements/deploy-replace-undeploy/package/REQ_get-op.adoc
+++ b/extensions/deploy_replace_undeploy/standard/requirements/deploy-replace-undeploy/package/REQ_get-op.adoc
@@ -3,7 +3,7 @@
 ====
 [%metadata]
 label:: /req/deploy-replace-undeploy/package/get-op
-part:: For every dynamically added process (path: /processes/{processId}), the server SHALL support the HTTP GET operation at the path `/processes/{processId}/package`.
+part:: For every dynamically added process (using method: POST on path: /processes/{processId}), the server SHALL support the HTTP GET operation at the path `/processes/{processId}/package`.
 part:: The parameter `processId` is each `id` property in the process collection response (JSONPath: `$.processes[*].id`).
 
 ====
--- a/extensions/deploy_replace_undeploy/standard/sections/clause_6_deploy_replace_undeploy.adoc
+++ b/extensions/deploy_replace_undeploy/standard/sections/clause_6_deploy_replace_undeploy.adoc
@@ -257,3 +257,28 @@ See <<deploy-replace-undeploy-http_status_codes,HTTP status codes>> for general 
 If the process with the specified identifier does not exist on the server, see requirement /req/core/process-exception/no-such-process from <<OAProc-1>>.
 
 If the process exist but is immutable then <<req_deploy-replace-undeploy_deploy_response-immutable,/req/deploy-replace-undeploy/deploy/response-immutable>> applies.
+
+[[deploy-replace-undeploy-package]]
+=== Retrieving the formal description
+
+For every process deployed to the server, it is possible to retrieve its formal description. It corresponds to the request's body used to deploy or replace the process.
+
+==== Operation
+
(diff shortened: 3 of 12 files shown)


extensions/deploy_replace_undeploy/standard/requirements/deploy-replace-undeploy/package/REQ_get-op.adoc
```diff
@@ -0,0 +1,9 @@
+[[req_deploy-replace-undeploy_package_get-op]]
+[requirement]
+====
+[%metadata]
+label:: /req/deploy-replace-undeploy/package/get-op
+part:: For every dynamically added process (path: /processes/{processId}), the server SHALL support the HTTP GET operation at the path `/processes/{processId}/package`.
```
fmigneault, line 6: Include `POST` along with `path: /processes/{processId}` to avoid ambiguity with `GET` from core.

extensions/deploy_replace_undeploy/standard/requirements/cwl/package/REQ_response-body.adoc
```diff
@@ -0,0 +1,10 @@
+[[req_cwl_package_response-body]]
+[requirement]
+====
+[%metadata]
+label:: /req/cwl/package/response-body
+part:: A response with HTTP status code `200` SHALL include a body that contains:
+ * the <<rc_cwl,CWL>> to use to deploy the process, in case the Content-Type used to deploy the process was `application/cwl`. 
+ * the <<rc_ogcapppkg,OGC Application Package>> to use to deploy the process, in case the Content-Type used to deploy the process was `application/ogcapppkg+json`.
```
fmigneault, line 8: Should we consider the possibility that a package deployed one way could provide the representation of another encoding? For example, if deployed using CWL, it would be fairly easy to extract the relevant parts of the document to form the corresponding *OGC Application Package* format. An `Accept` header could allow the selection of one, an if conversion is supported, could help make instances more interoperable. If so, sentences mentioning that "if deployed with X, SHOULD support X" would have to be reworded slightly.

Or, do we want to make it explicit that the original format for deployment must be returned (no conversion permitted)? In such case, the `Accept` header is irrelevant.

openapi/paths/processes-dru/pPackage.yaml
```diff
@@ -0,0 +1,25 @@
...
+              $ref: "../../schemas/processes-dru/ogcapppkg.yaml"
+         application/cwl:
+            schema:
+              $ref: "https://raw.githubusercontent.com/common-workflow-language/cwl-v1.2/1.2.1_proposed/json-schema/cwl.yaml"
+         application/cwl+json:
+            schema:
+              $ref: "https://raw.githubusercontent.com/common-workflow-language/cwl-v1.2/1.2.1_proposed/json-schema/cwl.yaml"
+         application/cwl+yaml:
+            schema:
+              $ref: "https://raw.githubusercontent.com/common-workflow-language/cwl-v1.2/1.2.1_proposed/json-schema/cwl.yaml"
```
fmigneault, line 23: The URI is not valid anymore (to fix in other parts of the spec as well). 
I already flagged the issue here: https://github.com/common-workflow-language/cwl-v1.2/issues/278#issuecomment-2079981922

Waiting for their feedback for an official location.

extensions/deploy_replace_undeploy/standard/requirements/cwl/package/REQ_response-body.adoc
jerstlouis: I vote yes, the implementation should be able to decide which package it can return back.

extensions/deploy_replace_undeploy/standard/sections/clause_6_deploy_replace_undeploy.adoc
```diff
@@ -257,3 +257,28 @@ See <<deploy-replace-undeploy-http_status_codes,HTTP status codes>> for general
 If the process with the specified identifier does not exist on the server, see requirement /req/core/process-exception/no-such-process from <<OAProc-1>>.
 
 If the process exist but is immutable then <<req_deploy-replace-undeploy_deploy_response-immutable,/req/deploy-replace-undeploy/deploy/response-immutable>> applies.
+
+[[deploy-replace-undeploy-package]]
+=== Retrieving the formal description
+
+For every process deployed to the server, it is possible to retrieve its formal description. It corresponds to the request's body used to deploy or replace the process.
```
pvretano, line 264: Perhaps this working is more clear?  "For every mutable process, it is possible to retrieve the application package that was used to deploy the process."

extensions/deploy_replace_undeploy/standard/requirements/cwl/package/REQ_response-body.adoc
gfenoy: I also favor having the capability to choose between various formats. I agree that CWL (and maybe other encoding. supported in the future) can be converted into an OGC Application Package and vice versa. Nevertheless, I wonder if the `Accept` header should not be added to the `6.6.1` section, which describes the operation for retrieving the formal description.

Then, in the `6.6 Retrieving the formal description` section, we would add another section: `Exception` specifying what happens if the `Accept` header is not supported. 

The `unsupported-media-type` exception type is possible if the media type is not supported. However, it's important to note that the exception could be something else if the process cannot be exposed in the desired encoding. For instance, consider a Docker image used to publish a process as an OGC Application Package, then a request `/processes/{processId}/package` using `Accept` header set to `application/cwl` should get back an exception that would help the client application determine that the issue comes from the fact that the conversion is impossible (maybe a type set to `http://www.opengis.net/def/exceptions/ogcapi-processes-2/1.0/incompatible-media-type`).

What do you think?

bpross-52n: SWG meeting from April 29th: Add a note that you can request the application package in a format other than originally used using the accept header. Examples: If `cwl` is explicitly requested and the server cannot handle it that a `HTTP 406` should be returned.

extensions/deploy_replace_undeploy/standard/sections/clause_6_deploy_replace_undeploy.adoc
fmigneault: The "mutable" process part raises a good question. What about immutable ones? I have some processes that are immutable, but still represented using CWL. The echo process is a good example of this, as it comes "pre-deployed" with the service.

pvretano: @fmigneault clients know which processes are mutable or not since there is a member in the process description that indicates that.  So, clients should not request the "package" of an immutable process since there is no guarantee that the "package" exists.  The process may, for example, be compiled right into the executable of the server.  
In the event that a client does ask for the package for an immutable process then they should be prepared to get a 404.  I am not saying that they will get a 404; if the server has the package or can generate it (as is your case), then the server can offer it up but in the event that a package does not exist the client should be prepared to get a 404.  At least that is was I think should happen.  Thoughts?

fmigneault: Yes. That is reasonable.

PR #405, opened 2024-04-29
Update definition of the f, prefer and {bbox-input-name}-crs parameters to minimize name collisions.

### LINKED ISSUE #402: Conflicting KVP execute parameters
## Description

As previously highlighted in #96 and #290, the KVP request format has potentially conflicting query parameters names.
Since there is no distinction between the `<inputID>=<value>` where `inputID` happens to be `prefer` or `f`, and control parameters `prefer=respond-async` or `f=application/json`, a process using those technically valid input IDs will be impossible to submit for execution. One potential solution would be to have `response[prefer]` and `response[f]` (or `response[format]`) to distinguish input IDs from control parameters.

Similarly, the bounding box input currently uses `<inputID>-crs` to provide the CRS. If the process took inputs with IDs `bbox` (BoundingBoxInput) and `bbox-crs` (LiteralStringInput), the execution request would cause side effects. There would be no way to distinguish from an input named `bbox-crs` and the "CRS for `bbox`". Instead, `bbox[crs]` should be used with the `deepObject` convention employed for all other formats (ie: by-reference, `output[include]` and so on). 

## References

- https://github.com/opengeospatial/ogcapi-processes/blob/1a54686aa27023ed6adb19c98dba3f55a39bb1d8/core/requirements/kvp-execute/REQ_prefer-definition.adoc
- https://github.com/opengeospatial/ogcapi-processes/blob/1a54686aa27023ed6adb19c98dba3f55a39bb1d8/core/requirements/kvp-execute/REQ_f-definition.adoc
- https://github.com/opengeospatial/ogcapi-processes/blob/1a54686aa27023ed6adb19c98dba3f55a39bb1d8/core/sections/clause_15_kvp-execute.adoc

pvretano:
Closes #402


core/requirements/kvp-execute/REQ_bbox-crs-input-value.adoc
```diff
@@ -12,13 +12,13 @@ identifier:: /req/kvp-execute/bbox-crs-input-value
...
-A bbox-crs input parameter SHALL have the following characteristics:
+A `inputCrs` input parameter SHALL have the following characteristics:
 
 [source,yaml]
 ----
 include::../../../openapi/parameters/common-crs/bbox-crs.yaml[]
 ----
 
-where the token `{bbox-input-name}` represents the identifier of a bbox input parameter and the token  `{bbox-input-name}-crs` represents the identifier of a corresponding input that indicates which CRS is being used for the coordinates of the `{bbox-input-name}` parameter.
+where the token `{bbox-input-name}` represents the identifier of a bbox input parameter and the token  `inputCrs[{bbox-input-name}]` represents the identifier of a corresponding input that indicates which CRS is being used for the coordinates of the `{bbox-input-name}` parameter.
```
fmigneault, line 22: Isn't it supposed to be the other way around? 
A bbox input named `test-bbox` would provide its CRS as `test-bbox[crs]=...`.
This would map well with `test-bbox[value]=...` for the qualified input value.
However, we imply that `test-bbox == test-bbox[value]` for ease of use.

In this case, the `inputCrs` name and the "variable" `{bbox-input-name}` are misleading. 

The text earlier should be "A `{bbox-input-name}` input parameter ..."

openapi/parameters/common-crs/bbox-crs.yaml
```diff
@@ -1,14 +1,20 @@
...
-    - http://www.opengis.net/def/crs/OGC/1.3/CRS84
-  default:
-    - http://www.opengis.net/def/crs/OGC/1.3/CRS84
-style: form
-explode: false
+  type: object
+  required:
+    - {bbox-input-name}
+  properties:
+    {bbox-input-name}:
```
fmigneault, line 12: Shouldn't this be a `additionalProperties`, or `patternProperties` with `.+[crs]`?
This definition would look for an explicit `{bbox-input-name}` property.

I think the solution would be: 

```yaml
in: query
name: inputs  # just a generic name to avoid ambiguity, ignored since 'explode: true'
explode: true
style: form
schema: 
  type: object
  patternProperties:
    "^.+[crs]$":
      type: string  
      format: uri  # etc. other params...
```

Ideally, we would have a separate schema for the following content, such that we can simply `$ref` to it under the pattern. https://github.com/opengeospatial/ogcapi-processes/blob/d011fe92750a0843106fcb87b6292db6e2b0fa51/openapi/schemas/processes-core/bbox.yaml#L16-L22

core/requirements/kvp-execute/REQ_bbox-crs-input-value.adoc
pvretano: Done.

openapi/parameters/common-crs/bbox-crs.yaml
pvretano: Done.

Issue #419, opened 2024-06-21
`POST /processes/{processID}/execution` is not RESTful and is confusing for workflows


m-mohr:
The standard says: 

> The standard specifies a processing interface to communicate over a RESTful protocol

In REST, everything should be centered around resources. The endpoint `POST /processes/{processID}/execution` is not a resource and POSTing to it, should create e.g. `/processes/{processID}/execution/{executionId}`, but it doesn't. Instead it may for example create `/jobs/{jobId}` (in async) (or return a result directly in sync).

To create a job, asynchronous processing requests should be sent to `POST /jobs`. This would also remove the issue that there is this weirdness that for workflows you need to pick a "main process" to work underneath.

Singular processes could also be sent there with a workflow that just consists of a single processing node.
if you just send async requests to the endpoint issues with the Prefer header would also be solved: #413 

Synchronous processing for a single process could still be sent to `POST /processes/{processID}/execution` but it would be weird for a workflow to be sent to that endpoint, too. So maybe it should be a separate top-level endpoint?

PS: This came up in some discussions with @fmigneault and @aljacob recently so posting it here for discussion.

fmigneault:
I will repeat my answer during the Testbed's meeting just for the sake of sharing with everyone openly. 

`POST /processes/{processID}/execution` was introduced (to my understanding), because an OAP implementation *is allowed* to omit the creation of a job. This is relevant, notably, if the OAP decides to *only* support sync execution mode, where a job resource is not necessary (though it could still create one for later reference if desired), since the results are obtained directly. 

Given that no job would be created in this case (which is technically considered the default/minimal requirement of OAP), the inverse non-RESTful argument arises if `POST /processes/{processID}/jobs` was used, since no job entity is created and 200 is returned. The way to avoid this ambiguity in REST is usually to replace the term by an action/verb, hence the `execution` (arguably, a better choice could have been `execute`?), to indicate that an *operation* is "created" rather than a resource.

Note that I agree with `/processes/{processID}/jobs` being better, since my implementation supports both sync/async, and creates a job for reference in both cases anyway, but I understand the reasoning of the execution counterpart. Since it is not much overhead, and for the sake of backward compatibility, my server handles both paths interchangeably.

[...]

PR #423, opened 2024-07-08
Add validation requirements class.

### LINKED ISSUE #101: Validating a process execution document without yet executing the process
In the context of building a bridge between our server supporting the _Workflows_ extension with a core _OGC API - Processes_, we realized it would be very useful to validate the submission of a process execution document without actually triggering the processing.

This could be as simple as introducing a "mode" : "validate", and would allow the server to validate all inputs, and if successful return the list of outputs (without actual data) that would have been returned if the process was executed. This may also be related to the _Quotation_ extension (e.g. "mode" : "quotation").

In the proposed Workflows extension, a workflow invovling one or more processes can first be validated along the full chain to make sure it will work, before requests can be issued for specific AoI+resolution. Supporting this would allow to implement this bridge while properly validating processes from APIs not supporting the _Workflows_ extension, used as part of such a workflow.

pvretano:
Closes #101.

### DIFF
--- a/core/requirements/validation/REQ_process-validate-op.adoc
+++ b/core/requirements/validation/REQ_process-validate-op.adoc
@@ -1,0 +1,8 @@
+[[req_process-validate_op]]
+[requirement]
+====
+[%metadata]
+identifier:: /req/process-validate/op
+
+The server SHALL support the HTTP POST operation at the path `/process/{processID}/validate`.
+====
--- a/18-062.adoc
+++ b/18-062.adoc
@@ -50,6 +50,8 @@ include::core/sections/clause_6_overview.adoc[]
 
 include::core/sections/clause_7_core.adoc[]
 
+include::core/sections/clause_7a_validation.adoc[]
+
 include::core/sections/clause_8_ogc-process-description.adoc[]
 
 include::core/sections/clause_9_security-considerations.adoc[]
--- a/core/requirements/requirements_class_validation.adoc
+++ b/core/requirements/requirements_class_validation.adoc
@@ -1,0 +1,12 @@
+[[rc_validation]]
+[requirements_class]
+====
+[%metadata]
+identifier:: http://www.opengis.net/spec/ogcapi-processes-1/1.0/req/validation
+obligation:: requirement
+subject:: Web API
+inherit:: http://www.opengis.net/spec/ogcapi_common-1/1.0/req/core[API - Common Core]
+inherit:: <<rfc2616,RFC 2616 (HTTP/1.1)>>
+inherit:: <<rfc2818,RFC 2818 (HTTP over TLS)>>
+inherit:: <<rfc8288,RFC 8288 (Web Linking)>>
+====
--- a/core/requirements/validation/REQ_process-validate-request.adoc
+++ b/core/requirements/validation/REQ_process-validate-request.adoc
@@ -1,0 +1,9 @@
+[[req_process-validate_request]]
+[requirement]
+====
+[%metadata]
+identifier:: /req/process-validate/request
+
+The content of a request the request body SHALL be based upon the OpenAPI
+3.0 schema https://raw.githubusercontent.com/opengeospatial/ogcapi-processes/master/openapi/schemas/processes-core/execute.yaml[execute.yaml].
+====
--- a/core/requirements/validation/REQ_process-validate-success.adoc
+++ b/core/requirements/validation/REQ_process-validate-success.adoc
@@ -1,0 +1,12 @@
+[[req_process-validate_success]]
+[requirement]
+====
+[%metadata]
+identifier:: /req/process-validate/success
+
+[.component,class=part]
+--
+A successful execution of the operation SHALL be reported as a response with a HTTP status code 200.
+--
+====
+
--- a/core/sections/clause_7a_validation.adoc
+++ b/core/sections/clause_7a_validation.adoc
@@ -1,0 +1,34 @@
+
+[[Validation]]
+== Requirements Class "Validation"
+
+=== Overview 
+
+include::../requirements/requirements_class_validation.adoc[]
+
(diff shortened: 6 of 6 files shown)


core/requirements/validation/REQ_process-validate-op.adoc
```diff
@@ -0,0 +1,8 @@
+[[req_process-validate_op]]
+[requirement]
+====
+[%metadata]
+identifier:: /req/process-validate/op
+
+The server SHALL support the HTTP POST operation at the path `/process/{processID}/validate`.
```
bpross-52n, line 7: should be /process/{processID}/validation

fmigneault:
IMO, we should focus on integrating the quotation and billing definitions that were already developed (though unpublished) with @pvretano over adding a new validation endpoint. The validation does essentially the same thing as requesting the quote using a pseudo-execution request, but quoting adds the additional estimate and ID for tracking of the original inputs submitted which can be used for the later execution without replicating the request/inputs again to actually run it.

I can dig up the old definitions we started working on. It looked something like the following:

OGC API - Processes - Quotation v2

OGC API - Processes - Billing v2


Other references: 
- OpenAPI: https://pavics-weaver.readthedocs.io/en/latest/api.html#tag/Billing-and-Quoting
- Docs: https://pavics-weaver.readthedocs.io/en/latest/quotation.html
- (some) Schemas: https://github.com/crim-ca/weaver/tree/master/weaver/schemas/quotation

pvretano:
@fmigneault @jerstlouis quotation and billiing is not something that would be in the core but rather something that would exist in another part of the suite of standards ... OGC API - Processes - Part X: Quotation and Billing.

Also, I am not sure I would explicitly tie the capability to validate a request to quotation and billing.  The first step in executing a request, or getting a quotation, is to validate the request.  All this PR does is simply expose an endpoint to allow that validation step to be performed independent of the next step (e.g. execution, quotation, etc.).  This is what Issue #101 is requesting.  If you now think this is not really necessary then I will delete this PR and close #101. 

We can create a new issue (although one might already exist) to create a new part of the standard dealing with quotation and billing.

jerstlouis:
@pvretano I would very much appreciate this "validate" thing in the core.

It's something that's very useful for workflows (to facilitate integratingprocesses implementing only Core in workflows), and it's something that _Part X: Quotation and Billing_ could extend with the extra capabilities.

I think we're saying that we should just think ahead in terms of Part X: Quotation and Billing integrating with this "don't actually execute the workflow" validating capability, in terms of end-points / parameters.

fmigneault:
I think it is acceptable to have this capability in Core. I agree it can be useful. Quotation can reuse the capability for the validation prior to the estimation. 

However, using `/processes/{processID}/validation` is still problematic for workflows. 
I just hope we don't get to the point where all those are needed: 

- `/processes/{processID}/validation`
- `/processes/{processID}/quotation`
- `/processes/{processID}/execution`
- `/workflow/{workflowID}/validation`
- `/workflow/{workflowID}/quotation`
- `/workflow/{workflowID}/execution`

jerstlouis:
My initial suggestion was a query parameter for the execution mode...

(sync / async), validation, quotation

hopefully cutting those end-points by 3 :)

If ad-hoc workflow execution is by POSTing a workflow to `/processes` with some query parameter indicting ad-hoc workflow execution rather than deployment then wouldn't need `/workflows/`.

fmigneault:
> ad-hoc workflow execution is by POSTing a workflow to /processes with some query parameter indicting ad-hoc workflow 

That is an interesting suggestion.
I think ah-hoc validation on `POST /processes` could make sense.
Not 💯 sold on the execution though. Might be more confusing when Part 2 is supported if the same endpoint that normally creates a process can also create a job for async. That will feel even less RESTful than the current `POST /processes/{processID}/execution`.

pvretano:
@fmigneault @jerstlouis nope!  I don't think overloading the `POST` method at the `/processes` endpoint so that one way it deploys a process and the other way it does workflow validation is a good idea at all!  

Again, a workflow is JUST another "execution unit" and I don't think "ad-hoc workflow execution" is a strong enough use case to upend everything just to to accomodate it ... at least not by overloading or redefining the current behaviour of the `/processes` endpoint.  We need to think about this a bit more!

fmigneault:
I agree with both points (not overloading the endpoint and no special treatment of ad-hoc over execution unit that handles all other use cases/implementations).

jerstlouis:
>  no special treatment of ad-hoc over execution unit that handles all other use cases/implementations

Right now, we have a "special treatment" in Part 1- Core for ad-execution of a single process, that is the POST to `/processes/{processId}/execution` with an execution request submitted as a payload, which customizes execution of the process with inputs.

Part 3: Workflows "Nested Processes" introduces the ability to execute ada-hoc workflows (chained from a root process) the same way.

I am really hoping to preserve all that functionality one way or another, for the ad-hoc execution use case.

It might not be a use case that everyone cares about, but I for one believe it is a very important use case, which facilitates discovery and integration of processes available from different deployment (where the delays associated with execution endpoints requiring credentials would be a barrier to experimenting with them in a timely manner).

gfenoy:
If we agree to use the POST on `/jobs` as proposed [here](https://github.com/opengeospatial/ogcapi-processes/issues/419#issuecomment-2216954782) to prepare a job, then I would be in favor of using GET on `/jobs/{jobid}/validation` and `/jobs/{jobId}/quotation`, and  POST on `/jobs/{jobId}/execution`.

pvretano:
@gfenoy @fmigneault @jerstlouis I do not agree!  I need more time to consider the issue but I am working on other things with the code sprint right now.  Let's discuss at the next SWG meeting.

bpross-52n:
SWG meeting from 2024-07-22: This PR will be closed and the validation endpoint will become part of a future extension.

PR #434, opened 2024-08-05
Clarify outputs

### LINKED ISSUE #372: outputs seem to be ambiguous
I'm writing a web UI for OGC API - Processes and I'm struggling with the outputs of processes.

Some assumptions
- An input for me is something where a user has a choice or can provide something to a process that usually changes the output ins some way. What the user can input is described in the schema.
- An output for me is what a process gives back and what is returned is described by the schema.

I compare this with math where f(x) = x*2 and f(2) returns 4. Are these assumptions true for OGC API - Processes?

Looking at some examples in the wild:

- https://testbed19.geolabs.fr:8713/ogc-api/processes/echo
- https://maps.gnosis.earth/ogcapi/processes/RenderMap?f=json

This is confusing about them:

1. In the RenderMap process there is an output and it defines that various media types can be returned. There's no parameter though that can influence the media type. The spec says content negotiation can be used, but the relation between content negotiation and the return type is sometimes there and sometimes isn't (see echo).
2. In the echo process it's not clear whether the process returns 3 values or whether the it's one output that can be any of the given output schemes. I'd assume it's one output with a choice of 3 types in RenderMap and 3 outputs wach with one type in echo. But then I'm also wondering, if I execute echo synchronously, how do I get the 3 types back in one response?

Overall, it looks like for RenderMap I need to render an additional select field for the file format in the UI, but for the echo process I don't. How can I distinguish that?

In the web UI I currently render an input for each output.
That works for Render Map:
grafik

But not very well for echo (which partially might also be a limitation in the UI):
grafik

### LINKED ISSUE #415: Sync / Multiple output response
[Requirement 31](https://docs.ogc.org/DRAFTS/18-062.html#_response_requesting_multiple_processing_outputs) could be adjusted to make it more obvious that alternative content types can be supported in this case. Right now it says:

> B	The media type of the response SHALL be application/json
> C	The content of response SHALL conform to the [results.yaml](https://raw.githubusercontent.com/opengeospatial/ogcapi-processes/master/openapi/schemas/processes-core/results.yaml) schema.

This is already somewhat conflicting with Permission 7 that says:

> Servers MAY support other response formats or encodings (e.g. ZIP or multipart/*) that do not conform to [results.yaml](https://raw.githubusercontent.com/opengeospatial/ogcapi-processes/master/openapi/schemas/processes-core/results.yaml).

I would suggest to not make it mandatory to support _results.yaml_, so as to allow sync-only servers to execute, return, forget as discussed in #412, without being forced to base64-encode large binary outputs, saying instead:

> B When the negotiated media type is `application/json`, the content of response SHALL conform to the [results.yaml](https://raw.githubusercontent.com/opengeospatial/ogcapi-processes/master/openapi/schemas/processes-core/results.yaml) schema.

and make it a recommendation instead to support `application/json` / _results.yaml_ content negotiation. Alternatively, if we want to keep the requirement to support application/json, the requirement could be rephrased as such to remove the apparent conflict with Permission 7:

> B The implementation SHALL support an `application/json` response
> C When the negotiated media type is `application/json`, the content of response SHALL conform to the [results.yaml](https://raw.githubusercontent.com/opengeospatial/ogcapi-processes/master/openapi/schemas/processes-core/results.yaml) schema.

In terms of the alternative encodings, we could also include GeoPackage (`application/geopackage+sqlite3`) as another excellent example of a media type supporting multiple outputs.

pvretano:
- Emphasizes that the `schema` in a process description input is the schema of a single instance value.
- Add a permission to say that when multiple output values are requested, something other than `result.yaml` (e.g. ZIP, multi-part MIME) can be used as long as the client and server negotiate to that.

Closes #415
Closes #372


pvretano:
05-AUG-2024:  We will merge this PR because, in addition to addressing issues #372 & #415, it removes a lot of kruft and fixes some cross reference issues with the recommendation.  If there are further comments on the PR, we can open new issues.

Issue #464, opened 2024-10-24
Part 4: Job Provenance endpoint mismatches GDC API Profile


fmigneault:
Part 4 currently proposes using `GET /jobs/{jobId}/run` for provenance (https://github.com/opengeospatial/ogcapi-processes/tree/master/extensions/job_management/standard/requirements/provenance/run).

In TB-20 GDC, the `GET /jobs/{jobId}/prov` endpoint is considered instead.

Given that the contents returned by `GET /jobs/{jobId}/run` are expected to be PROV specification (https://github.com/opengeospatial/ogcapi-processes/blob/master/extensions/job_management/standard/requirements/provenance/run/REQ_response.adoc), it makes sense to use `GET /jobs/{jobId}/prov`, and leave `GET /jobs/{jobId}/run` for other metadata (if any). 

Should the endpoint be changed, or an alias be defined?

gfenoy:
If we find anything that should be considered, we can re-open this issue or open another issue.

Issue #455, opened 2024-10-19
Part 4: Clarify /req/job-management/definition-get-op


m-mohr:
>  What is the definition meant to return? The process? (e.g. CWL or openEO UDP?)
> 
> We include that directly in GET /jobs/:id - I guess that's fine and we can add an additional endpoint, but maybe this is more an optional endpoint for cases where a definition can't be embedded? And it can be explored via a link in GET /jobs/:id?
> 
> What is POST /jobs/:id?

_Originally posted by @m-mohr in https://github.com/opengeospatial/ogcapi-processes/pull/437#discussion_r1789875331_

gfenoy:
The idea was initially to return the definition used to instantiate the job (`"status": "created"`) with additional metadata to describe the mode used (sync/async), for example. Still, I think it would be better to store it in the `process` attribute (see #450), which would then be added to the `statusInfo.yaml` schema, and we can remove this endpoint.

What do you think?

m-mohr:
I agree. Default should be the `process` property. We can define an optional link relation type for processes that can't be represented in JSON, but without a specific endpoint.

fmigneault:
Because jobs are submitted using a `process: URI` property, I wouldn't be surprised that some servers were already embedding that process URI in the job status response. Therefore, the property added to `statusInfo.yaml` should not only consider the embedded JSON representation, but the reference directly as well (using `oneOf`).

m-mohr:
OAP jobs might be, openEO no. There it's `process: object`. And what is if you submit MOAW or CWL? That's not necessarily URI either, right?

fmigneault:
Exactly. This is why a separate endpoint (or the same using another Accept header) is proposed. This we, we don't need to depend on a specific embedding of `process` within the job (though I'm not against having it embedded if the server provides it). Just need to allow the flexibility of what `process` contains.

m-mohr:
Yeah, that's what I proposed. That doesn't necessarily need a separate pre-defined endpoint though. Not sure whether we agree or disagree right now 😅

fmigneault:
If `process: "https://.../processes/{processId}"` can be returned in the job status, and that endpoint supports `Accept` header to negotiate any `application/cwl+json`, `application/ogcapppkg+json`, etc. handled by the server, then yes, `GET /jobs/{jobId}/definition` is redundant.

fmigneault:
[...]

PR #473, opened 2024-11-14
Feature/fix issue 464

gfenoy:
As described in #464, the `/job/{jobId}/run` endpoint initially proposed should be renamed to `/job/{jobId}/prov` as per ongoing work in Tesbed 20 GDC activities.

We also added a `headers` array containing the specific headers used to create the job, as described in a not here: https://github.com/opengeospatial/ogcapi-processes/issues/455#issuecomment-2450618499.


openapi/schemas/processes-job-management/inputs.yaml
```diff
@@ -19,3 +19,7 @@ properties:
     type: array
     items:
       $ref: "../processes-core/metadata.yaml"
+  headers:
+    type: array
+    items:
+      $ref: "headers.yaml"
```
fmigneault, line 25: Can this be made directly `$ref: "headers.yaml"`?

PR #477, opened 2024-12-05
Changes for closer alignment with OpenEO.

pvretano:


openapi/schemas/processes-core/statusCode.yaml
```diff
@@ -1,8 +1,13 @@
 type: string
 nullable: false
 enum:
-  - accepted
+  - created
+  - queued
   - running
+  - cancelled
```
m-mohr, line 7:
```suggestion
  - canceled
```

openEO uses "canceled", is it double l in OGC API Processes?

openapi/schemas/processes-core/statusInfo.yaml
```diff
@@ -1,42 +1,47 @@
...
+          $ref: process.yaml
+       type:
+          type: string
+          example:
+            - process
+            - wps
+            - openeo
+       status:
+          oneOf:
+          $ref: "statusCode.yaml"
```
m-mohr, line 23:
```suggestion
       status:
          oneOf:
          $ref: "statusCode.yaml"
```

Looks invalid

openapi/schemas/processes-core/statusInfo.yaml
```diff
@@ -1,42 +1,47 @@
...
-      items:
-         $ref: "../common-core/link.yaml"
+allOf:
+  - $ref: descriptionType.yaml
+  - type: object
+    required:
+       - id
+       - status
+       - created
+       - process
```
m-mohr, line 8: Should it be required or optional?

openapi/schemas/processes-core/statusInfo.yaml
pvretano: Fixed.

openapi/schemas/processes-core/statusInfo.yaml
pvretano: I reverted back to the required fields for OAProc and included text in the descriptions based on the value of the "type" property.  If the value of "type" is "openeo" then both "created" and "process" MUST be present as per the OpenEO API definition.

openapi/schemas/processes-core/statusInfo.yaml
```diff
@@ -1,42 +1,47 @@
...
+       - type
+    properties:
+       id:
+          type: string
+       process:
+          $ref: process.yaml
+       type:
```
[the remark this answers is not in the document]
pvretano: I've updated the description to indicate that if "type" is "openeo" then the "process" property MUST be present and its value shall be a process graph as per the OpenEO API definition.  Otherwise if can be an OAProc process description, a WPS process description, a link to either of these or a link to the request that created the job (e.g. an async coverages request).

openapi/schemas/processes-core/statusInfo.yaml
```diff
@@ -1,42 +1,99 @@
...
+       - id
+       - type
+       - status
+    properties:
+       id:
+          type: string
+       type:
+          type: string
+          description:
+            An indication of the API that created this job.
```
fmigneault, line 14: You will need a `|` to do the multiline string and keep the newlines.

Same thing for all following `description` fields.

openapi/schemas/processes-core/statusInfo.yaml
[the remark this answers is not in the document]
fmigneault: Do we want this condition to be more explicit?

Since we have the `allOf`, we can make use of it.

```yaml
allOf:
  - $ref: descriptionType.yaml
  - type: object
    # ... all other fields ...
  - if:
      properties:
        type: 
          const: openeo  # or 'enum' equivalent if prefered
    then:
      required:
        - id
        - type
        - status
        - process
      properties:
        process:
          $ref: "<uri-to-openeo-process-graph>"
```

openapi/schemas/processes-core/statusInfo.yaml
```diff
@@ -1,42 +1,99 @@
...
+          oneOf:
+            - $ref: "statusCode.yaml"
+            - type: string
+              enum:
+                - created
+                - queued
+                - running
+                - canceled
+                - finished
+                - error
```
fmigneault, line 71: Does openEO have a schema for this?
The definition could be more explicit if we actually pointed to `$ref: <uri-openeo-schema#status>` or wherever it resides.

openapi/schemas/processes-core/statusCode.yaml
```diff
@@ -3,6 +3,6 @@ nullable: false
 enum:
   - accepted
   - running
-  - succeeded
+  - successful
```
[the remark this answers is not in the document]
pvretano: I looked at the 1.0 schema [here](https://schemas.opengis.net/ogcapi/processes/part1/1.0/openapi/schemas/statusCode.yaml) and it uses "successful" so I figured we should use that.  We will need to propagate that change everywhere in the 2.0 version.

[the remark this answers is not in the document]
fmigneault: Works for me.

openapi/schemas/processes-core/statusInfo.yaml
m-mohr: We don't do the OGC way with snippets, we only have a monolithic OpenAPI document. In principle you can refer to that though:

`"$ref:" "https://raw.githubusercontent.com/Open-EO/openeo-api/refs/tags/1.2.0/openapi.yaml#/components/schemas/batch_job/properties/status"`

openapi/schemas/processes-core/statusInfo.yaml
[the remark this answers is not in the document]
m-mohr: I'm wondering whether OGC API - Processes should just define the framework and not have direct dependencies to openEO due to the issue that there are multiple versions of the openEO API...

As long as openEO follows your framework, I guess it's fine, isn't it?

openapi/schemas/processes-core/statusInfo.yaml
m-mohr: Thanks. 

Generally, I think we can do better to avoid these if/then conditionals as much as possible :-)

openapi/schemas/processes-core/statusInfo.yaml
fmigneault: The second link would be ideal. It's fine with definition anchors. Either YAML or JSON should work.

openapi/schemas/processes-core/statusInfo.yaml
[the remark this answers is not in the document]
fmigneault: I'm expecting that anyone that wants to do strong schema validation would actually use the reference openEO schemas rather than depend on OGC API, so it's fine if we don't define field requirements of openEO explicitly here. 

However, if we do so, I don't see much point to have the long description either for specific frameworks. Saying in the description that `process` is required when `type: openeo` doesn't actually enforce or validate anything.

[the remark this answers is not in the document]
m-mohr: Yes, I think the spec should not include specifics about openEO.

PR #480, opened 2025-01-05
Add link for log files and update status info.

### LINKED ISSUE #273: Logging feature?
Short: as a developer of processing (geo-)routines, I wish I could share with the clients some logs through specific endpoint(s) of the OGC API - Processes in order to share these important pieces of information with them.

This is maybe a naive question but I cannot figure out an answer in the draft standard document: is the "OGC API - Processes" featuring some logging abilities or ability to collect and return some processing logs, _separate_ from the results per se, e.g. for sharing intermediate states, information on particular processing sub-steps, errors, timestamps, etc., i.e, whatever they may be (without having to specify for each individual log feature a specific optional property)?

pvretano:
Closes #273


bpross-52n:
SWG meeting from 2025-01-06: Peter will add a definition of workflow.

openapi/schemas/processes-core/execute.yaml
```diff
@@ -1,5 +1,7 @@
 type: object
 properties:
+  processID:
+    type: string
```
fmigneault, line 4: To align with other parts that use the URI to allow nested processes reference (Part 3) and POST'ing directly on `/jobs` (Part 4), `process: <uri>` would be preferable.

openapi/schemas/processes-core/statusInfo.yaml
```diff
@@ -1,42 +1,48 @@
...
+  - type: object
+    required:
+      - id
+      - status
+      - type
+    properties:
+      id:
+        type: string
+      processID:
+        type: string
```
fmigneault, line 12: Similarly to previous comment. The job's process might not reside on this server. Therefore, `process: <uri>` is preferable. Consider also that `process` might be an object at some point (openEO alignment), similar to the `request` property below.

openapi/schemas/processes-core/execute.yaml
pvretano: Done.

openapi/schemas/processes-core/statusInfo.yaml
pvretano: Done.

pvretano:
> SWG meeting from 2025-01-06: Peter will add a definition of workflow.

Decided to remove the reference to "workflow" and simply used the more generic "request" which can be an execute request or a workflow.

PR #499, opened 2025-06-16
adjust description of `process` in execute-workflows.yaml

fmigneault:
The URI description is misleading when looking at examples (eg: https://docs.ogc.org/DRAFTS/21-009.html#_192ed695-ac52-4c60-97a1-c2df43bd7ac9). 

It points at the process *description* endpoint, not the execution.


fmigneault:
@pvretano @jerstlouis

pvretano (approved):

PR #509, opened 2025-06-27
Add missing `Accept-Profile` for Content Negotiation by Profile

fmigneault:
The standard's clause for *Content Negotiation by Profile* (https://github.com/opengeospatial/ogcapi-processes/blob/33f775d298cf2d8eed57b872f5828222dcf340ae/core/sections/clause_7_core.adoc#content-negotiation-by-profile) currently mentions 3 alternatives that matches definitions under https://www.w3.org/TR/dx-prof-conneg/#related-http

However, one of the key element of `profile` parameter provided within `Accept` / `Content-Type` headers is that they are "technically" only valid "__*if the media type includes provisions for such a parameter*__", as already mentioned by the above clause.

Given that, reading more into the subtext of the section [Content Negotiation by Profile](https://www.w3.org/TR/dx-prof-conneg/#using-accept-content-type-together-with-the-profile-parameter), one can read:

> *In order to signal the use of Media Type-independent profiles, the http header fields `Accept-Profile` and `Link: <...>; rel="profile"` are to be used.*

Therefore, even if the https://profilenegotiation.github.io/I-D-Profile-Negotiation/I-D-Profile-Negotiation is yet to be established, the `Accept-Profile` is __*already part*__ of the [Content Negotiation by Profile](https://www.w3.org/TR/dx-prof-conneg/#using-accept-content-type-together-with-the-profile-parameter) reference used to define this section.

Furthermore, the description was missing some detail about which headers must be returned, given that a certain strategy of profile negotiation is employed and resolved.

All of this is added to this PR

### DIFF
--- a/core/sections/clause_7_core.adoc
+++ b/core/sections/clause_7_core.adoc
@@ -262,13 +262,24 @@ include::../recommendations/core/REC_html.adoc[]
 
 The use of generic media types such as `application/json` makes it difficult to request or identify different flavors or profiles of a resource.  For example, the media type `application/json` could be used to describe an <<ogc_process_description,OGC process description document>> as well as an https://documentation.dataspace.copernicus.eu/APIs/openEO/Processes.html[OpenEO process description document].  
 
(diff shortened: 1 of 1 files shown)


fmigneault:
@pvretano Feel free to edit/adjust the wording as desired.

PR #531, opened 2026-01-12
[OAPROC CODE SPRINT JAN-2026] Feature/no such output (Close #529)

### LINKED ISSUE #529: Add an exception for invalid or unknown outputId (CODE SPRINT JAN-2026)
Since the standard already has an exception for an invalid {jobID} (a path element) we should also add an exception for and invalid {outputID} which is also an invalid path element.

gfenoy:
This update is related to issue #529 we created during the code sprint.

### DIFF
--- a/core/requirements/core/REQ_job-results-exception-no-such-output.adoc
+++ b/core/requirements/core/REQ_job-results-exception-no-such-output.adoc
@@ -5,8 +5,24 @@
 identifier:: /req/core/job-results-exception-no-such-output
 label:: /req/core/job-results-exception-no-such-output
 
-If the endpoint is requested with an invalid output identifier, the response SHALL have an HTTP status code of 400.
-The content of that response SHALL be based upon the OpenAPI
-3.0 schema https://raw.githubusercontent.com/opengeospatial/ogcapi-processes/master/openapi/schemas/common-core/exception.yaml[exception.yaml].
+
+[.component,class=conditions]
+--
+.The endpoint is requested using an invalid output identifier `outputID`.
+--
+
+[.component,class=part]
+--
+The response SHALL have HTTP status code 400.
+--
+
+[.component,class=part]
+--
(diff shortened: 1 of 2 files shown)


fmigneault (changes requested):

core/abstract_tests/core/ATS_job-results-exception-no-such-output.adoc
```diff
@@ -0,0 +1,31 @@
...
+[%metadata]
+identifier:: /conf/core/job-results-exception-no-such-output
+target:: /req/core/job-results-exception-no-such-output
+test-purpose:: Validate that the job results retrieved using an invalid output identifier complies with the require structure and contents.
+test-method::
++
+--
+1. Issue an HTTP GET request to the URL '/jobs/{jobID}/results/{outputID}' using an invalid {outputID}.
+
+2. Validate that the document was returned with a 404.
```
fmigneault: I would consider "invalid" to be the case of "not aligned with the process description's outputs" (like your `.../results/d` example). Therefore, 400 would apply. 

The 404 / 410 would be for other cases, such as:
- 404: output does not exist because the process execution was not requested with it
  (that case would still not generate the result if re-submitting the execution as is)
- 410: output does not exist *anymore*, although it was originally present in the execution result 
  (that case would allow re-submitting the execution as is to obtain it)

As discussed, a server would be permitted to use 404 instead of 410 if it does not want to keep track of previous results, but 400 would indicate that this ID never works regardless.

core/requirements/core/REQ_job-results-exception-no-such-output.adoc
```diff
@@ -0,0 +1,12 @@
+[[req_core_job-results-exception-no-such-output]]
+[requirement]
+====
+[%metadata]
+identifier:: /req/core/job-results-exception-no-such-output
+label:: /req/core/job-results-exception-no-such-output
+
+If the endpoint is requested with an invalid output identifier, the response SHALL have an HTTP status code of 404.
```
fmigneault: Same as other comment.

bpross-52n:
SWG meeting from 2026-01-19: We will check the versioning of the exception type.

Issue #538, opened 2026-01-15
Feedback on 2.0 (CODE SPRINT JAN 2026)


jerstlouis:
Below is my feedback reading through the candidate standard (up to end of 7-Core -- will create separate issue for Sections 8 Collections+):

**Abstract**

- The Abstract has tables. I think it would be a better fit to move these to Section 6 Overview which is currently very small. AFAIK the Abstract is intended to be text only / easy to include in a summary description of the document
- I would also suggest not including so much "background context" about WPS in the abstract (other than e.g., one sentence that OGC API - Processes is the modern REST/JSON-based successor to WPS), and focus more on summarizing what this particular standard is. Maybe the more detailed WPS background could be a sub-section of the Overview section
- In table 1, I would call the "Result" resource "Individual output result" to clarify the distinction with "Job results"
- "Additionally, the `/jobs` endpoint can be used to grant access to a list of jobs." -- I do not understand what this sentence means

**Preface**

The Standard specified **who** to “wrap” computational tasks into an executable process that can be invoked by a client application.

_How_?

**Security Section**

- It's a bit unusual that the Recommendation 1 is within the Security section rather than within the main body of the document
- Everywhere in the document it seems that there is both Identifier and Label (with identical URIs). I think we should only keep
 Identifier (Label was the pre-metanorma thing).

**Acknowledgements**
- Please include an acknowledgement section to credit the work done as part of our GeoConnections project using the following exact language below (and other things if appropriate):
> Requirements integrated within version 2.0 of this OGC API - Processes Standard was sourced from Part 3: Workflows, whose development was undertaken as part of GeoConnections 2020-2021 and 2025-2027 projects GNS20-21IFP02 (Modular OGC API Workflows) and GNS25-27IFP2 (Advancing and Implementing OGC Standards for GeoDataCubes, DGGSs, Data Stores for Digital Twins and Cartographic Symbology). Financial support provided by GeoConnections, a national collaborative initiative led by Natural Resources Canada. GeoConnections supports the modernization of the Canadian Geospatial Data Infrastructure (CGDI). The CGDI is the collection of geospatial data, standards, policies, applications, and governance that facilitate its access, use, integration, and preservation.

**Conformance**

- for the resources specified in the Core**a** (remove the a)
- The Collection Input, Remote Collections, Collection Outout, Local Filtering conformance classes are missing here from both the table and summary.
[...]

PR #543, opened 2026-01-19
fix typos in KVP schemas

fmigneault:
- fix typos in KVP schemas
- Add the "query" as suggested in https://github.com/opengeospatial/ogcapi-processes/issues/538

### DIFF
--- a/core/requirements/core/REQ_job-results-param-outputs.adoc
+++ b/core/requirements/core/REQ_job-results-param-outputs.adoc
@@ -7,7 +7,7 @@ label:: /req/core/job-results-param-outputs
 
 [.component,class=part]
 --
-The operation at the `/jobs/{jobID}/results` endpoint SHALL support a parameter `outputs` with the following characteristics (using an OpenAPI Specification fragment):
+The operation at the `/jobs/{jobID}/results` endpoint SHALL support a query parameter `outputs` with the following characteristics (using an OpenAPI Specification fragment):
 
 [source,yaml]
 ----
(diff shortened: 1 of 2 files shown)


fmigneault:
I am fairly sure this was not intended as the `outputs` of the execution request, since that one is already detailed in great lengths. 

The idea is that you could process an entire product once (e.g.: the 13 bands of a sentienl-2 image), and then extract various band subset combinations (RGB, RGB-NIR, etc.), or even do further processing with them (e.g.: NVDI), just by reusing the filtered job result with them, e.g.: using `/jobs/{jobID}/results?outputs=B04,B03,B02&f=geotiff`, if the process happens to have them produced separately (as for STAC Assets for example). 

However, since that can quickly become equivalent to invoking another process execution, defining a workflow, or doing a mixture of `?response=collection` with `?properties=B04,B03,B02` with other OGC APIs, I think it is fine to not further expand on it in Core. It is simply another possible option to filter the results, and you can then do whatever you want with them according to content-negotiation.

jerstlouis:
@fmigneault 

> The idea is that you could process an entire product once (e.g.: the 13 bands of a sentienl-2 image), and then extract various band subset combinations

But what I was saying is that this `outputs` query parameter does not help at all for that use case, since you could still:

- Request each output band that you need separately as `/results/{outputId}`, or
- Request `/results` which is just a small JSON file including the link to each individual output (likely as links to `/results/{outputId}`) and then only follow the ones you're interested in

It's only useful for requesting `/results` as `application/zip` or other media type that could really contain multiple outputs together, or when specifying `Prefer: return=representation` and trying to get huge outputs returned inside a huge single JSON results file.

PR #546, opened 2026-02-02
First cut of input and output bindings in the execution unit.

pvretano:


pvretano:
@fmigneault I agree that the "closer" it looks like CWL then a server implementation should just use CWL.  

However, since OGC APP PKG is still in the standard and it is missing this one important bit that allows the mapping from process input to command line then we should add something and since CWL has already done the hard work of figuring out what is needed, it makes sense to steal and adapt this one little bit of CWL into OGC APP PKG.

Eventually we will deprecate OGC APP PKG and remove it altogether if no one is using it.

Thanks for the other feedback ...

PR #547, opened 2026-02-16
Changes to address issues raised during the JAN-2026 OAProc Code Sprint

### LINKED ISSUE #530: Add permission to allow HTTP 410 status code (CODE SPRINT JAN2026)
In general, if a resource once existed but no longer exists because it was expired our or deleted, the server would return a 404 in this case if the resource is subsequently requested. 
We should add a permission to the relevant sections of the standard indicating that the server could optionally return a 410 Gone.
From the client perspective, this lets the client know that they did not make a mistake with their request; rather the resource once existed and was valid but is now gone (either deleted or expired out).

### LINKED ISSUE #532: Requirements 48,49 and 50 have an extra path element (CODE SPRINT JAN-2026)
The above-listed requirements have an extra path element ("job-results-exception") that is not allowed by the mod spec.  This extra path element should be removed or the slash replaced by a dash.

Also, requirement 48 seems too generic to be only for job results.  We should move that somewhere else or remove it altogether.  Actually we should copy what features does [here](https://docs.ogc.org/is/17-069r4/17-069r4.html#query_parameters) and say that if an undefined or invalid query parameter is specified, or the value of the parameter is not valid then the server should thrown a 400.

### LINKED ISSUE #534: passThroughParameters (CODE SPRINT JAN 2026)
Right now we have a [`passThruParameters` property](https://docs.ogc.org/DRAFTS/18-062r3.html#sc-pass-thru).

It is currently called a _parameter_ but I think it is more accurate to call this a **property** (of a collection input object).

In the text, including the sections headers (that show up in the table of content) it is described as "Pass thru parameters".
Since this is [not standard spelling](https://www.merriam-webster.com/dictionary/thru) we should fix this to `through`.

For the property, having the short/bad orthograph only saves 3 letters of what is already 18 letters long. So I think we should either spell it out completely `passThroughParameters` or rename it to simply `passThrough`. 

What do you think @pvretano @fmigneault ?

### LINKED ISSUE #538: Feedback on 2.0 (CODE SPRINT JAN 2026)

### LINKED ISSUE #539: Remove the "respond-sync" value (CODE SPRINT JAN 2026)
Right now the standard seems to have two way of requesting synchronous execution.  One way is based on the "wait" preference of the Prefer header.  The other way is to use the "respond-sync" preference which is a preference defined in the OAProc standard but not in the "Prefer Header of HTTP" standard.

We should probably remove the unofficial  "respond-sync" preference.

That, however, puts an extra burden on clients since there would no longer be a deterministic way to request synchronous execution for processes that support both "sync-execute" and "async-execute".

I suppose that if a server had a process "BOB" that supports both "sync-execute" and "async-execute" and the server wants to also offer a deterministic way to execute "BOB" synchronously, the server could advertise another identical process (called "BOB-SYNC" for example) that only support "sync-execute".

### LINKED ISSUE #540: Feedback on 2.0 / Collections (CODE SPRINT JAN 2026)
Requirement 53:

- that contain the collection _parameter_  -- can we say _property_ instead of parameter here (B also below)
-  **process input values** are retrieved -- isn't this somewhat confusing to call this "input values"? We could say something like for "which data is retrieved for a specific input"? Since we have the concept of "process inputs", and a collection input retrieves data (bunch of values) for one specific input.
- C --  The SHALL here is not written from the server's perspective, and offers too much flexibility regarding relative URIs. Suggesting:
> The server SHALL support both collection identifiers (for collections hosted on the same API as the Processes API) and absolute URIs of the collection description (but not sub-resources of the collection). This allows the server to decide, based on the expected input value type from the [process description](https://docs.ogc.org/DRAFTS/18-062r3.html#sc_process_description) and other factors, which supported [data access mechanisms](https://docs.ogc.org/DRAFTS/18-062r3.html#def-data-access-mechanism) should be employed to retri**e**ved input values from the [collection](https://docs.ogc.org/DRAFTS/18-062r3.html#def-local-collection). (note the typo in retri**e**ved)
- Suggesting we get rid of D in favor of the collectionIds option above
- Req. 54: retri**e**ved typo again
- 8.1.3.2.  Filtering -- We should add some informative text here that clarifies that with the use of Collection Output / Virtual Collections, these filtering / subsetting parameters would often NOT be specified in the execution requests, and instaed the server would automatically subset/filter as a direct result of what is being requested from the output/virtual collection. 
- Requirement 56 (derived fields) -- This should be its own requirements/conformance class. Typos: `dervied` -> `derived`, `correspoding` missing an **n**.
- 8.2.2.  Referencing a remote collection -- _contain the collection parameter._ collection **property** (in B also). C: Similar comment: The server SHALL support absolute URIs of a collection description (but not sub-resources of the collection). 
- 8.3.1.  Overview -- `prameters` missing an **a**.   _would need to be continuously **be** re-submitted._ -- extra `be`
- 8.3.2.  Parameter response -- clarify it's a **query** parameter (also in Req. 60 A)
- Permission 13:  Condition 5, 6 `profiles` query parameter should be singular?

### LINKED ISSUE #541: Feedback on 2.0 (continued -- CODE SPRINT JAN 2026)
- 9.3.2.  Data classes dataClasses _parameter_ -- can we please change this to `property` as well (and anywhere _parameter_ is used for a meaning other than _query_ or _path_ parameter)
- 9.3.3 69 C: typo in coverages:  ogc-api-covr**e**ages (remove the `e`)
- There are conflicting spellings in the informative text s. the requirement -- should it be plural with uppercase API? Would make sense to me that it is plural `dataAccessAPIs` (it's singular and `Api` in the normative clauses)
- The example has messed up metanorma rendering
- The messed up example has typos param**a**ter and accomo**n**date
- Req. 70 schema parameter -- schema **property**; B: inpu**i**t value typo
- Req. 72 Value passing B and D shall is not in uppercase. Use **property** rather than parameter here again for `valuePassing`
- Req. 72 -- it took me a while to understsand what is meant by the 4 clauses of this requirement... Perhaps it would be good to start with something like:

> If a server implementation only supports input values passed either by reference or by value, the `valuePassing` property of the process description SHALL list the supported input passing mechanism.

It's also very odd to call this property `valuePassing` with "byValue" being one of the option. How about `inputPassing` instead?
- 9.3.7 A: _correspoding_ typo missing `n`
- `executionUnitRequrements` parameter -> property
- `executionUnitRequrements` -- it's not clear what `remote-access` and `staging` are -- are these sub-properties of the `executionUnitRequrements` property, which is a top-level property of the process description? also `remote-access` should probably be camelCase rather than kebab-case.
- 9.3.7 -- it's not clear to me whether this pertains to Part 1 or not? Shouldn't this rather be in Part 2? Does it have any implications on how a client submits an execution request? Or is it more about how the Part2-capable server handles execution units which expect to handle networking etc.?
- 9.4.2.  Data classes parameter and 9.4.3.  Data access APIs have some TBD (but also See Also)
- Req. 80 /req/oas/definition -- This seems to have kept some Features API-specific language about Feature Collections that should be adapted to processes or generalized
- Req. 89, it's confusing to talk about "the parameter" (**type query parameter** in this case, which is a reference to Req 88) without context because it is in a completely different requirement. In general, I'm not a fan of splitting request/response into different requirements because I feel it makes the documents longer than they need to be (I prefer to use different A, B parts instead), but if the requirement is split, there should be enough context for a reader arriving directly at any of the requirements and make sense of it (e.g., being pointed there due to an ETS failure)
- `datetime` query parameter -- that should probably talk about bounded/half-bounded intervals rather than open/closed as was clarified before in OGC API - Features etc.

### LINKED ISSUE #542: KVP requirements not using new metanorma templates (CODE SPRINT JAN 2026)
15 - KVP: the requirements seem to still be using the pre-metanorma templates

### LINKED ISSUE #545: Add input binding section of OGC APP PACKAGE ... (CODE SPRINT JSN 2026)
The one thing missing from the OGC APP PACKAGE description is the binding between the input description in the process description and the binding to the execution unit (e.g. docker).  This needs to be added to make is easier for implementation to generate the command line to execute in something like a docker container.  The suggestion from the recent code is to copy/mirror the input binding from CWL.

pvretano:
Closes #530
Closes #532
Closes #534
Closes #538
Closes #539 
Closes #540 
Closes #541
Closes #542
Closes #545


fmigneault:
@pvretano 
Is this mostly ready to go? It would help to work based on https://github.com/opengeospatial/ogcapi-processes/tree/master with the latest changes.

core/recommendations/core/REC_process-execute-handle-prefer.adoc
```diff
@@ -1,9 +0,0 @@
-[[rec_core_process-execute-handle-prefer]]
-[recommendation]
-====
-[%metadata]
-identifier:: /rec/core/process-execute-handle-prefer
-label:: /rec/core/process-execute-handle-prefer
```
fmigneault, line 6: This is removed because a requirement elsewhere?

Issue #564, opened 2026-03-19
Do we really want to allow qualifying array elements INSIDE an input/output occurrence?


jerstlouis:
@pvretano 

There was this change made in Core where [valueNoObject.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/valueNoObject.yaml) now takes:
```yaml
  - type: array
    items:
      oneOf:
        - $ref: "inputValue.yaml"
        - $ref: "qualifiedValue.yaml"
```

instead of (old [valueNoObject.yaml](https://schemas.opengis.net/ogcapi/processes/part1/1.0/openapi/schemas/inputValueNoObject.yaml)):
```yaml
  - type: array
    items: { }
```

Why? Would you know which issue number discussed this, if any?

Sorry I was not able to follow all the discussions with everything going on.

The way that [input.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/input.yaml) is structured, it already does:

```yaml
oneOf: 
  - $ref: "inlineOrRefValue.yaml" 
  - type: array 
    items: 
      $ref: "inlineOrRefValue.yaml" 
```

So this is what accounts for the multiplicity and qualification of the input values.

So what you would have in [inlineOrRefValue.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/inlineOrRefValue.yaml) is the input value (individual occurrence) itself.
What this change does now is let you "qualify" things INSIDE the input value (individual occurrence, and in a recursively nested fashion), clashing with the specified JSON Schema in the process description.
This change does not seem right to me.
Before, that generic array was whatever array described in the process description... It's not an input that should be qualifiable or including valueNoObject.yaml again...
That might have something to do with the deep recursion that was breaking the older Swagger UI (I had to migrate to the newer version to get over these errors).

cc. @fmigneault @gfenoy @bpross-52n

fmigneault:
It **forces** (not just "let") you to use the qualified input structure if you want to pass an array (cardinality / `minOccurs>=1`) of complex objects. 

Before this change, there is no way to distinguish complex structures that happens to have a `value` property vs a qualified one, and it makes parsing highly nested structure pure guesses (i.e.: along which nested array/object you can safely scale out the operation).

jerstlouis:
[...]

PR #566, opened 2026-03-27
Fixes for #564 (full alignment of outputs / inputs schemas with maxOccurs)

jerstlouis:
This implements the proposed fixes for #564 

(please focus on the last commit c426de7f8511e688914a7eee5282e0e08c4960ef , as this PR is otherwise based on #561 which fixes the validation of the OpenAPI definition)


pvretano:
As a native speaker of English I can tell you that the name "occurrencesHolder.yaml" is a meaningless name.  Please rename the file to "inputOrResultValue.yaml".

jerstlouis:
@pvretano We can discuss what the proper name is, but this name was intended to carry the meaning of "something that holds occurrences" (of an input or result output value) :) Just like a cup holder holds a cup ;)

@fmigneault was strongly objecting to the use of `input` in the name...

I still think using `value` in the name is confusing because it might be multiple values (multiple occurrences), and all of the other YAML files with value in the filename are about individual value occurrences (not the holder / shell / input / output / result)

pvretano:
@jerstlouis change the name.  It is meaningless and the term occurrence is not used once in the standard.  If it can be one or more values then use the plural "values".

jerstlouis:
@pvretano 

> change the name.

What to? `inputOrResultValues.yaml` ?

> the term occurrence is not used once in the standard.

Independently of what this file is called, in my opinion it would be very useful in the standard to talk about _one or more occurences of values_ (as per min**Occurs** / max**Occcurs**) in the context of the inputs and result outputs to clarify this lingering confusion between the two, explaining it a bit like I did in [this comment](https://github.com/opengeospatial/ogcapi-processes/issues/564#issuecomment-4121059645) (EDIT: Sorry I had linked to the wrong comment). This is certainly confusing readers/implementers, but also ourselves in the SWG (as per the JSON Schema clash issue identified by #564).

I suggest to use this as potential inspiration:

There is the input/output which is:
- If maxOcccurs = 1: `<occurrence>`
- If maxOccurs > 1:` [ <occurrence>, ... ]` (empty array allowed as well unless minOccurs > 0)

Where `<occurrence>` is one of:
- `{ "value" : ... }`
- `{ "href" : ... }`
- `{ "bbox" : ... }`
- `{ "collection": ... }`
- a string
- a number
- `true` or `false`
- an array
- `null`

There is no confusion about an object defining a `"value"` property at the top level. That would always be encoded as:
`{ "value" : {"value": 123 } }`.

pvretano:
@jerstlouis inputOrResultsValues.yaml.

jerstlouis:
@pvretano Now there is an issue with `results` having an s and input not having one.

This file is about one or more (or 0) values for an individual input or result.

So I'll rename it to `inputOrResultValues` with no s at result.

pvretano:
@jerstlouis fine...

jerstlouis:
@pvretano Done. Thanks. See 5a666d03f14e4ffb8c5ac4c60166598d149956d0 .

fmigneault:
So is it a good time to say that, including *Part 3: Workflows*, `inputOrResultValues.yaml` now is ambiguous, since it should be `inputOrResultValues-OrProcess-OrParamerized.yaml` because they can be nested?

`inputOrResultValues-workflows.yaml` is not that much more communicating the variants.

(my reasoning for making it abstract, but I didn't want to bring up too many extra concepts simultaneously until the first batch of renames were addressed)

jerstlouis:
@fmigneault 

A nested process or a parameterized input (`$input`) would still be an input value to the process that receives it, so the name still fits.

The only thing is that [inputParameterized](https://github.com/opengeospatial/ogcapi-processes/blob/fix564/openapi/schemas/processes-workflows/inputParameterized.yaml) and [inputOrResultValues-workflows](https://github.com/opengeospatial/ogcapi-processes/blob/fix564/openapi/schemas/processes-workflows/inputOrResultValues-workflows.yaml) are only used for inputs, not for results, so we could potentially rename that one to simply `inputValues-workflows.yaml`, or keep like this to more easily see the relationship to the no-workflows version.

fmigneault:
Good point. Fair enough. It highlights even more why they must be the same "value container" structure in this case!

pvretano:
Guys, are you serious now?  We don't need long sentence like-file names!  Even "inputOrResultValues.yaml" is too long but it is a sight better than "occuranceHolder.yaml" ... What is this?  Dungeon & Dragons!?  

Just call the file "values.yaml" and let the context dictate whether it is an input value, a result value, a nested value in a workflow or whatever.  I want these PR's either merged or discarded on Monday so that I can complete my PRs so that we can get on with the RFC.

fmigneault:
> Just call the file "values.yaml" and let the context dictate whether it is an input value, a result value, a nested value

I agree with this. A simple `values.yaml` is what I've proposed since the start.

jerstlouis:
@fmigneault @pvretano If you both want me to rename [inputOrResultValues.yam](https://github.com/opengeospatial/ogcapi-processes/blob/fix564/openapi/schemas/processes-core/inputOrResultValues.yaml) to `values.yaml` in this PR and that will help agreeing to merge this PR tomorrow I'm happy to change it.

As long as we all remember (from this epic Dungeons & Dragons adventure ;)) that `values.yaml` is the shell / holder / container, while all the other files with _value_ in them are for the individual values occurring zero, once, or multiple times inside that container.

pvretano:
@jerstlouis make the change.

jerstlouis:
@fmigneault @pvretano Done. Thanks!

pvretano (approved):
As a native English speaker I can tell you that the name "occurencesHolder.yaml" is completely meaningless.  The term "occurrence" is not used in the text of the standard.  Please rename the file to "inputOrResultValue.yaml" (or Values if plural makes more sense).

Issue #567, opened 2026-03-31
OpenAPI: Missing definition of the 2.0 `/jobs/{jobId}/results/{outputId}`


jerstlouis:
The OpenAPI definition is still missing that path.

I am happy to prepare a PR for this.

I think that should reference the [`values.yaml` ](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/values.yaml) schema for responses negotiated with `application/json`.

We should also discuss whether we want to add a requirement or recommendation for `/jobs/{jobId}/results/{outputId}/{N}`.

(related to #414 discussiong maxOccurs > 1)

Also related to #527 filed by @gardengeek99 , since I think this is one of the major changes for 2.0 that is missing (other than the newer decision to move the collection input/output stuff to Part 1).

cc. @fmigneault @pvretano

Issue #560, opened 2026-03-19
`processID` vs. `process` in execute.yaml


jerstlouis:
@pvretano 

This commit 157a26733d0f7d42f8808cbd5fa7fb5e48ae687d added back a `processID` with a format URI in [https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/execute.yaml](execute.yaml).

Assuming they are more or less the same thing, this would be conflicting with the `process` used in Part 3: Workflows (see [execute-workflows.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-workflows/execute-workflows.yaml)), or otherwise lead to confusion.

For a workflow definition, it's more expressive to simply say "process" than "processID", since that nested process input objected is actually an execution of a process, not an abstract reference to a "process ID".

I'm not sure why this `processID` was added to Core -- that commit message says "_Add link for log files and update status info_" and has a lot of changes, but I suggest that this simply be the same `process` URI as Workflows if needed.

fmigneault:
duplicates https://github.com/opengeospatial/ogcapi-processes/issues/514

jerstlouis:
@pvretano Can I also correct and merge a fix for this issue and the #514 duplicate that @fmigneault opened in August last year?

pvretano:
This is a process identifier ... is it not?  If yes then it should be "processID" not "process".  The term "process" is too vague in my opinion.  So leave it as is in part 1 and change it in part 3.

jerstlouis:
@pvretano In the context of Part 3: Workflows, this can be a full URI to a process (and the process ID can be valid substitute for a local process).

While technically you could say the process ID or the URI is an "identifier", it really represents the process itself being executed.

Just like we call e.g., `https://example.com/collections/myCollection` a collection rather than a collectionId.

And this also aligns with the Collection Input which is `{ "collection": "https://example.com/collections/myCollection" }` rather than `"collectionId"`, and the collectionID can be a valid substitute for a local collection. So in the case of Part 3, it is a "Process Input".

We discussed this previously in 2021 in #123 among several other issues I think.

[...]

Issue #414, opened 2024-05-26
Add minOccurs and maxOccurs to output description


pvretano:
Right now, the specification makes the assumption that the correlation between a named output and an output value is 1:1.  However, I've run into use cases where it would be convenient to allow the correlation to be 1:N or some other ratio.  I'm thinking of a simple use case where you have a process that has a single input with `minOccurs=1` and `maxOccurs=unbounded` and generates a single output ... but you want one output value for each specified input.  

Right now the only way I can see this being described is using the `schema` element with an array but that kinda breaks the "the scheme describes a single instance value" rule that applies to inputs.

Adding `minOccurs` and `maxOccurs` to the output description and extending the results path to `/jobs/{jobId}/results/{outputID}/[N]` where N is the Nth output value that corresponds to the Nth input value would allow a more symetrical description on inputs and outputs.

The same rules would apply to fetching a single output or multiple outputs that currently apply.  If you `GET /jobs/{jobId}/results/{outputID|` you get a results.yaml with all the output values.  If you fetch `/jobs/{jobId}/results/{outputID}/[N]` you get the content-negotiated output as you currently do.  The defaults would be set to be backward compatible.

pvretano:
Might be related to #412.

jerstlouis:
> Right now the only way I can see this being described is using the schema element with an array but that kinda breaks the "the scheme describes a single instance value" rule that applies to inputs.

If the output definition is a schema defined as an array of 1 or more elements, with a description that says there will be as many elements in that array as occurences of that input, I don't think this breaks anything -- the schema does define one output: an array of multiple elements making up the single output instance.

> If you `GET /jobs/{jobId}/results/{outputID}` you get a results.yaml with all the output values. 

Having a _results.yaml_ response at this end-point would introduce so much additional confusion, because _results.yaml_ is used for `/jobs/{jobId}/results` when negotiating `application/json`.

Things remain much simpler if it is only used there, because normally `GET /jobs/{jobId}/results/{outputID}` always returns the actual output. There was already so much confusion about the meaning of _minOccurs_ and _maxOccurs_ with the input, that I think we should avoid introducing it for the outputs.

But if one wants to do something like you describe, then one could simply define a JSON schema for the output that's similar to _results.yaml_ with links to individual outputs? (the links could probably even be to `/jobs/{jobId}/results/{outputID}/[N]` since that doesn't seem to break anything).

Other potential alternative way to handle such use cases:
[...]

PR #568, opened 2026-03-31
OpenAPI fixes for #414 #560 #567

jerstlouis:
- Adding missing paths `/jobs/{jobID}/results/{outputID}` and `/jobs/{jobID}/results/{outputID}/{N}`
- Change `processID` to `process` in Core to match workflows and the fact that this is a full URI
- Change "collection" from `uri-reference` to `uri` or `string` (`{collectionId}`) as discussed in https://github.com/opengeospatial/ogcapi-processes/issues/540#issuecomment-3774139649
- Changing `"process": ` in Workflows to allow `uri` (absolute URI to the process) or `string` (`{processId}`) instead of `uri-reference` (unchanged in Part 1, since process should probably always be a full URI in this case since it is at the top level)

### DIFF
--- a/openapi/schemas/processes-workflows/execute-workflows.yaml
+++ b/openapi/schemas/processes-workflows/execute-workflows.yaml
@@ -4,7 +4,18 @@ allOf:
         process:
           type: string
           format: uri-reference
-          description: URI to the process description end point (i.e., `.../processes/{processId}`)
+          description: |-
+             The URI to the description of the process that should be executed.
+             For the top-level process, when the execution request is POSTed to `{root}/processes/{processId}/execution`, this should be an absolute URI,
+             and the server should ignore this `process` key since the URI of the request already identifies the process to execute, and a client
+             is not required to resolve and replace the URI if passing a nested process block as part of a larger workflow executing a nested process.
+             If a relative URI reference is used, the Base URI is defined as the URI of the description
+             (`{root}/processes/{processId}`) of the receiving process, with this receiving process object understood as the
+             Encapsulating Entity as per [RFC 3986 Section 5.1.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.2).
+             In the case of a nested process on the same OGC API root as the receiving process, using a simple process ID as the relative URI reference
+             would resolve to a process available at `{root}/processes/{processID}`, where `{root}` represents the landing page of the receiving process.
+             For example, if a top-level process at https://example.com/processes/buffer receives a nested processs `{ "process": "clip" }`,
+             that clip process is resolved as https://example.com/processes/clip.
         inputs:
           additionalProperties:
             $ref: "values-workflows.yaml"
--- a/openapi/ogcapi-processes.bundled.json
+++ b/openapi/ogcapi-processes.bundled.json
@@ -630,9 +630,6 @@
           },
           {
             "$ref": "#/components/parameters/N"
-          },
-          {
-            "$ref": "#/paths/~1jobs~1%7BjobID%7D~1results/get/parameters/1"
           }
         ],
         "responses": {
--- a/openapi/ogcapi-processes.bundled.json
+++ b/openapi/ogcapi-processes.bundled.json
@@ -5989,7 +5989,7 @@
               "process": {
                 "type": "string",
                 "format": "uri-reference",
-                "description": "URI to the process description end point (i.e., `.../processes/{processId}`)"
+                "description": "The URI to the description of the process that should be executed.\nFor the top-level process, when the execution request is POSTed to `{root}/processes/{processId}/execution`, this should be an absolute URI,\nand the server should ignore this `process` key since the URI of the request already identifies the process to execute, and a client\nis not required to resolve and replace the URI if passing a nested process block as part of a larger workflow executing a nested process.\nIf a relative URI reference is used, the Base URI is defined as the URI of the description\n(`{root}/processes/{processId}`) of the receiving process, with this receiving process object understood as the\nEncapsulating Entity as per [RFC 3986 Section 5.1.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.2).\nIn the case of a nested process on the same OGC API root as the receiving process, using a simple process ID as the relative URI reference\nwould resolve to a process available at `{root}/processes/{processID}`, where `{root}` represents the landing page of the receiving process.\nFor example, if a top-level process at https://example.com/processes/buffer receives a nested processs `{ \"process\": \"clip\" }`,\nthat clip process is resolved as https://example.com/processes/clip."
               },
               "inputs": {
                 "additionalProperties": {
@@ -6183,8 +6183,17 @@
             "properties": {
               "collection": {
                 "description": "The URI of the OGC API collection that should be accessed to provide\ninput values from the corresponding process input.  The server\noffering this collection is referred to as the \"value source server\".",
-                "type": "string",
-                "format": "uri-reference"
+                "anyOf": [
+                  {
+                    "type": "string",
+                    "description": "Collection identifier for a collection available at `{root}/collections/{collectionID}`, where `{root}` represents the landing page\nof the process using the collection as an input (for the top-level process, that is the landing page of the process execution endpoint\nto which the execution request is posted), or the landing page of the document being retrieved in the case of an output.\nIf the `collection` string includes a `/`, it must be interpreted as a URI reference rather than as a collection ID."
+                  },
+                  {
+                    "type": "string",
+                    "description": "The URI of an OGC API collection.\nIf a relative URI reference is used, the Base URI is defined as such:\n- For a Collection value received as input to a process, the Base URI is defined as the URI of the description\n  (`{root}/processes/{processId}`) of that process, with the receiving process object understood as the Encapsulating Entity as per\n  [RFC 3986 Section 5.1.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.2).\n- For a Collection Output value, the Base URI is defined from the Retrieval URI as as per\n  [RFC 3986 Section 5.1.3](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.3),\n  for example, `/jobs/{jobId}/results`, `/jobs/{jobId}/results/{outputID}` or `/jobs/{jobId}/results/{outputID}/{N}` depending on the\n  resource being retrieved.\nIf a collection string does not contain any `/`, it must not be interpreted as a URI reference, and must instead be resolved as a\ncollection identifier. For example, `countries` would resolve to `{root}/collections/countries` rather than `{root}/processes/countries`.",
+                    "format": "uri-reference"
+                  }
+                ]
               },
               "bbox": {
                 "description": "Only resources that have a geometry that intersects the bounding box\nare selected.  The bounding box is provided as four or six numbers,\ndepending on whether the coordinate reference system includes a\nvertical axis (height or depth):\n\n* Lower left corner, coordinate axis 1\n* Lower left corner, coordinate axis 2\n* Minimum value, coordinate axis 3 (optional)\n* Upper right corner, coordinate axis 1\n* Upper right corner, coordinate axis 2\n* Maximum value, coordinate axis 3 (optional)\n\nIf the value consists of four numbers, the coordinate reference\nsystem is WGS84 longitude/latitude\n  (http://www.opengis.net/def/crs/OGC/1.3/CRS84)\nunless a different coordinate reference system is specified in the\nparameter `bboxCrs`.\n\nIf the value consists of six numbers, the coordinate reference system\nis WGS84 (http://www.opengis.net/def/crs/OGC/0/CRS84h) longitude/\nlatitude/ ellipsoidal height unless a different coordinate reference\nsystem is specified in the parameter `bboxCrs`.\n\nFor WGS84 longitude/latitude the values are in most cases the sequence\nof minimum longitude, minimum latitude, maximum longitude and maximum\nlatitude.  However, in cases where the box spans the antimeridian the\nfirst value (west-most box edge) is larger than the third value\n(east-most box edge).\n\nIf the vertical axis is included, the third and the sixth number are\nthe bottom and the top of the 3-dimensional bounding box.\n\nIf a resource has multiple spatial geometry properties, it is the\ndecision of the server whether only a single spatial geometry property\nis used to determine the extent or all relevant geometries.",
--- a/openapi/ogcapi-processes.bundled.json
+++ b/openapi/ogcapi-processes.bundled.json
@@ -580,6 +580,73 @@
           }
         }
       }
+    },
+    "/jobs/{jobID}/results/{outputID}": {
+      "get": {
+        "summary": "retrieve a specific resulting output of a job",
+        "description": "Retrieve a specific resulting output for a job.\n\nFor more information, see [Section 7.14](https://docs.ogc.org/is/18-062r3/18-062r3.html#sc_retrieve_job_results_one).\n",
+        "operationId": "getOutput",
+        "tags": [
+          "Jobs"
+        ],
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/jobID"
+          },
+          {
+            "$ref": "#/components/parameters/outputID"
+          },
+          {
+            "$ref": "#/paths/~1jobs~1%7BjobID%7D~1results/get/parameters/1"
+          }
+        ],
+        "responses": {
+          "200": {
+            "$ref": "#/components/responses/Output"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound"
+          },
+          "500": {
+            "$ref": "#/components/responses/ServerError"
+          }
+        }
+      }
+    },
+    "/jobs/{jobID}/results/{outputID}/{N}": {
+      "get": {
+        "summary": "retrieve a single value from an output of a job allowing multiple values",
+        "description": "Retrieve a single value from an output of a job allowing multiple values (maxOccurs > 1)\n\nFor more information, see [Section 7.14](https://docs.ogc.org/is/18-062r3/18-062r3.html#sc_retrieve_job_results_one).\n",
+        "operationId": "getOutputValue",
+        "tags": [
+          "Jobs"
+        ],
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/jobID"
+          },
+          {
+            "$ref": "#/components/parameters/outputID"
+          },
+          {
+            "$ref": "#/components/parameters/N"
+          },
+          {
+            "$ref": "#/paths/~1jobs~1%7BjobID%7D~1results/get/parameters/1"
+          }
+        ],
+        "responses": {
+          "200": {
+            "$ref": "#/components/responses/Output"
+          },
+          "404": {
+            "$ref": "#/components/responses/NotFound"
+          },
+          "500": {
+            "$ref": "#/components/responses/ServerError"
+          }
+        }
+      }
     }
   },
   "components": {
@@ -7304,6 +7371,15 @@
           "type": "string"
         }
       },
+      "N": {
+        "name": "N",
+        "in": "path",
+        "description": "0-based index identifying a specific value of a multi-values output resulting from a job",
+        "required": true,
+        "schema": {
+          "type": "string"
+        }
+      },
       "jobID": {
         "name": "jobID",
         "in": "path",
@@ -7628,6 +7704,26 @@
           "application/json": {
             "schema": {
               "$ref": "#/components/schemas/results"
+            }
+          }
+        }
+      },
+      "Output": {
+        "description": "An output resulting from a job",
+        "content": {
+          "application/json": {
+            "schema": {
+              "$ref": "#/components/schemas/values"
+            }
+          }
+        }
+      },
+      "OutputValue": {
+        "description": "An individual value of an output allowing multiple values resulting from a job",
+        "content": {
+          "application/json": {
+            "schema": {
+              "$ref": "#/components/schemas/inlineOrRefValue"
             }
           }
         }
--- a/openapi/ogcapi-processes.yaml
+++ b/openapi/ogcapi-processes.yaml
@@ -59,6 +59,11 @@ paths:
    /jobs/{jobID}/results:
       $ref: 'paths/processes-core/pJobResults.yaml'
 
+   /jobs/{jobID}/results/{outputID}:
+      $ref: 'paths/processes-core/pResultOutput.yaml'
+
+   /jobs/{jobID}/results/{outputID}/{N}:
+      $ref: 'paths/processes-core/pResultOutputValue.yaml'
 
 components:
    schemas:
@@ -439,9 +444,12 @@ components:
       # Processes - Core
       processID-path:
          $ref: 'parameters/processes-core/processID-path.yaml'
-      # Processes - Core (1.1?)
+      # Processes - Core (2.0)
       outputID:
          $ref: 'parameters/processes-core/outputID.yaml'
+      # Processes - Core (2.0)
+      N:
+         $ref: 'parameters/processes-core/N.yaml'
 
       # Processes - Core (Jobs)
       jobID:
@@ -503,6 +511,10 @@ components:
 
       Results:
          $ref: 'responses/processes-core/rResults.yaml'
+      Output:
+         $ref: 'responses/processes-core/rOutput.yaml'
+      OutputValue:
+         $ref: 'responses/processes-core/rOutputValue.yaml'
       Status:
          $ref: 'responses/processes-core/rStatus.yaml'
       JobList:
--- a/openapi/parameters/processes-core/N.yaml
+++ b/openapi/parameters/processes-core/N.yaml
@@ -1,0 +1,6 @@
+name: N
+in: path
+description: 0-based index identifying a specific value of a multi-values output resulting from a job
+required: true
+schema:
+  type: string
--- a/openapi/paths/processes-core/pResultOutput.yaml
+++ b/openapi/paths/processes-core/pResultOutput.yaml
@@ -1,0 +1,21 @@
+get:
+   summary: retrieve a specific resulting output of a job
+   description: |
+     Retrieve a specific resulting output for a job.
+
+     For more information, see [Section 7.14](https://docs.ogc.org/is/18-062r3/18-062r3.html#sc_retrieve_job_results_one).
+
+   operationId: getOutput
+   tags:
+     - Jobs
+   parameters:
+     - $ref: "../../parameters/processes-core/jobID.yaml"
+     - $ref: "../../parameters/processes-core/outputID.yaml"
+     - $ref: "../../parameters/processes-core/prefer-header-results.yaml"
+   responses:
+     200:
+       $ref: "../../responses/processes-core/rOutput.yaml"
+     404:
+       $ref: "../../responses/common-core/rNotFound.yaml"
+     500:
+       $ref: "../../responses/common-core/rServerError.yaml"
--- a/openapi/paths/processes-core/pResultOutputValue.yaml
+++ b/openapi/paths/processes-core/pResultOutputValue.yaml
@@ -1,0 +1,22 @@
+get:
+   summary: retrieve a single value from an output of a job allowing multiple values
+   description: |
+     Retrieve a single value from an output of a job allowing multiple values (maxOccurs > 1)
+
+     For more information, see [Section 7.14](https://docs.ogc.org/is/18-062r3/18-062r3.html#sc_retrieve_job_results_one).
+
+   operationId: getOutputValue
+   tags:
+     - Jobs
+   parameters:
+     - $ref: "../../parameters/processes-core/jobID.yaml"
+     - $ref: "../../parameters/processes-core/outputID.yaml"
+     - $ref: "../../parameters/processes-core/N.yaml"
+     - $ref: "../../parameters/processes-core/prefer-header-results.yaml"
+   responses:
+     200:
+       $ref: "../../responses/processes-core/rOutput.yaml"
+     404:
+       $ref: "../../responses/common-core/rNotFound.yaml"
+     500:
+       $ref: "../../responses/common-core/rServerError.yaml"
--- a/openapi/paths/processes-core/pResultOutputValue.yaml
+++ b/openapi/paths/processes-core/pResultOutputValue.yaml
@@ -12,7 +12,6 @@ get:
      - $ref: "../../parameters/processes-core/jobID.yaml"
      - $ref: "../../parameters/processes-core/outputID.yaml"
      - $ref: "../../parameters/processes-core/N.yaml"
-     - $ref: "../../parameters/processes-core/prefer-header-results.yaml"
    responses:
      200:
        $ref: "../../responses/processes-core/rOutput.yaml"
--- a/openapi/responses/processes-core/rOutput.yaml
+++ b/openapi/responses/processes-core/rOutput.yaml
@@ -1,0 +1,5 @@
+description: An output resulting from a job
+content:
+  application/json:
+    schema:
+      $ref: "../../schemas/processes-core/values.yaml"
--- a/openapi/responses/processes-core/rOutputValue.yaml
+++ b/openapi/responses/processes-core/rOutputValue.yaml
@@ -1,0 +1,5 @@
+description: An individual value of an output allowing multiple values resulting from a job
+content:
+  application/json:
+    schema:
+      $ref: "../../schemas/processes-core/inlineOrRefValue.yaml"
--- a/openapi/schemas/processes-core/collectionValue.yaml
+++ b/openapi/schemas/processes-core/collectionValue.yaml
@@ -9,8 +9,27 @@ allOf:
            The URI of the OGC API collection that should be accessed to provide
            input values from the corresponding process input.  The server
            offering this collection is referred to as the "value source server".
-        type: string
-        format: uri-reference
+        anyOf:
+           - type: string
+             description: |-
+               Collection identifier for a collection available at `{root}/collections/{collectionID}`, where `{root}` represents the landing page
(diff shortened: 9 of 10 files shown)


pvretano (approved):

fmigneault:
The change of `uri-reference` introduces a problem where (not technically, but in the wording) it disallows a relative URI, since "absolute" is indicated for the URI while the other is only for `{root}/processes/{processID}`. The `uri-reference` allowed either variant including `../processes/other`. 

I think `Prefer: return=minimal|representation` makes sense for single output. One of the main reason why a single-output would be requested individually rather than the full set of results is precisely to get an alternate representation. Eg: the process returns all results by default `return=minimal` (i.e.: links) to keep it small, but you then want to retrieve one of those links directly as data (e.g.: offer a preview of one result) using `Prefer: return=representation`.

jerstlouis:
@fmigneault 

> The change of uri-reference introduces a problem where (not technically, but in the wording) it disallows a relative URI, since "absolute" is indicated for the URI while the other is only for {root}/processes/{processID}. The uri-reference allowed either variant including ../processes/other.

There is no change regarding this for the execute.yaml of Part 1.

I only changed the execute.yaml for Part 3, which was intentional, and the collection inputs. Absolute URI or `{processId}`. Anything else is very ambiguous as to what is the context to which it is relative and an unnecessary burden on implementations to support. We discussed this together at length in the context of collections in #540, and I thought we had agreed to Absolute URIs or IDs, but avoiding the term _local_ which you found confusing and instead using the landing page / `{root}/processes` approach.

> One of the main reason why a single-output would be requested individually rather than the full set of results is precisely to get an alternate representation.

I removed the `Prefer:` header for a single **value** of a  single output, but not for a single output of the results.

fmigneault:
It is not ambiguous. It is relative to the endpoint where the request is sent, just like any other HTTP link resolution. Both `my-process` and `../processes/my-process` are prone to the exact same problem: if sent as-is anywhere else than the same `{root}`, their interpretation change.  It is up to the API to resolve those links as it receives the contents and make sure it is not ambiguous for the rest of the downstream operations. I don't have a problem to allow `{processId}` for URI or ID convenience, but it is not a reason to disallow relative URIs formed as `../processes/my-process`.

---

From my understanding, the only difference between `../outputs/out` and `../outputs/out/1` is the nested array element on multiple occurrence. Therefore, if the default minimal result is like so:
```yaml
out:
  - href: https://...
    type: application/json
  - href: https://...
    type: application/json
```

Both  `../outputs/out` and `../outputs/out/1` could use `Prefer: return=representation`. The first would return every element of the array as `value: {json data}` (qualified or directly), whereas the `../outputs/out/1` would return just the `N=1` single `value: {json data}` extracted from the array.

jerstlouis:
@fmigneault 

>  It is relative to the endpoint where the request is sent, just like any other HTTP link resolution. Both my-process and ../processes/my-process are prone to the exact same problem

In the context of Part 1, there is only one top-level process, and that is the one whose `/execution` end-point you POST to. So a `{processID}` or a relative ID are both useless for the POST operation, _unless_ it is an absolute URI, which is useful e.g., for GDAL to be able to open an execution request, concatenate `{processURI}` + `/execution`, and treat the outputs as a data source.

The other case which might be why this "process" was added could have been to retrieve the execution of the process. In this case, the execution request would be a GET operation to somewhere like `/jobs/{jobId}/execution` or `/collections/{collectionID}/workflow` or similar? Then the content of the execution request would not be exactly as submitted, since the base URL changes, which poses an additional significant headache.

In the context of Part 3, a nested `{ "process": ... }` would be relative not to where you POST the execution request to, but to the process using that process object as an input. This is what allows a workflow engine to only work one level down, and send an entire nested input within `{ curlies }` to the immediate hop, with the ability to ignore everything within (it could of course decide to validate what it sends down to the next process, but it is not a technical necessity, other than for security reasons).

Regarding `../processes/my-process`, I'm not sure that what is relative to, as if you consider a POST to `/processes/SomeProcess/execution` as the base URL, wouldn't that yield `/processes/processes/my-process`?

In any case, I don't see why we would require servers to support arbitrary relative URIs, when the only thing useful is a `{processID}` or a full URI, and the relative URIs are in fact ambiguous because the reference is the receiving process rather than the execution end-point, and the execution request needs to be modified from where it was POSTed to where it is retrieved.

---

> Both ../outputs/out and ../outputs/out/1 could use Prefer: return=representation. The first would return every element of the array as value: {json data} (qualified or directly), whereas the ../outputs/out/1 would return just the N=1 single value: {json data} extracted from the array.

I'm struggling to follow.
Let's assume the output has maxOccurs > 1.
First we should clarify whether `{N}` is using base-0 indexing or base-1 (I added the parameter with a description saying it is base 0 indexing).
There's also no `/outputs`, it's `/results/{outputID}` (as confusing as that is).

If maxOccurs > 1, `GET /jobs/{jobId}/results/{outputID}/0` for `application/json`, will return [inlineOrRefValue.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/inlineOrRefValue.yaml). (EDITED)

If maxOccurs > 1, `GET /jobs/{jobId}/results/{outputID}` would be one or more links to the data with `Prefer: return=representation`, and an array with base64-encoded values with `Prefer: return=representation`, with an ability to qualify each value in both cases. For `application/json`, the response is always [values.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/values.yaml).

EDIT: `Prefer: return=minimal` returning only the `{ "href": ... }` object would have some very limited utility  for `GET /jobs/{jobId}/results/{outputID}/0`, as `GET /jobs/{jobId}/results/{outputID}` could return all links at once. In fact, if you were to return only a link, you would probably return a link to itself saying `{ "href": "/jobs/{jobId}/results/{outputID}/0" }`, assuming the output is stored within that server's root.

fmigneault:
I agree with you that, for the underlying algorithm/program (eg: GDAL or other), the absolute URI is relevant. However, this is something for the OGC API that receives the request to deal with. If the API deems it is relevant to pass down the process URI to the program for it to work correctly, it should be doing any necessary absolute URI resolution beforehand. From a functional point of view, submitting `../processes/my-process` (e.g.: to `/jobs`) should resolve to the exact same location, just like `{processID}` should be translated to the absolute URI for the same GDAL operation. Therefore, there is no reason an API request couldn't handle that relative URI prior to passing it down to the program, since the program should always receive the pre-resolved process reference. 

An advantage of relative URI is that I could send the same request to multiple deployments offering the same set of processes without having to care about the specific hostname for each (notably since they could be dynamically created and not known it advance), and I still get the flexibility to refer to multiple *OGC API - Processes* deployments living side by side (in other to refer to various `{root}`). It would allow automatically scaling that execution request with many servers without need to manually edit the execution contents for each host variation. That makes sense if you have to send requests to many places to compute close to the relevant data location.

> In the context of Part 3, a nested { "process": ... } would be relative not to where you POST the execution request to, but to the process using that process object as an input. 

IMO, that just adds confusion. It seems to work against how the internet works. Use URIs relative to the request context they are sent to. If you want references to nested elements within the body (eg: the parent `process` reference), it should be `$refs`, `$id`, `$anchor` or another similar relation hint, not working against the HTTP linking mechanisms that are all relative to the __*request context*__ (i.e.: just like an `href: ../` in `links` would point to the parent URI location). 

If you want to refer to a parent process under the same `{root}`, you can already do it with a plain `{processID}`. You don't need another abstract reference way to do it. I'm not really sure to understand your reason about the workflow engine, or why a nested process would need any further relative reference to the parent, as it is already nested within that parent, so all context is already there.  It seems specific to your implementation or a specific representation you have on your end. IMO, it is entirely up to the API and workflow engine to agree on these references, and apply any optimization/resolution however it wants. The REST API request receiving that request should do any necessary work to interpret the contents __*from a request context*__ in a way that is appropriate for the __*workflow engine context*__ to do its job. If that means substituting URIs or `{ curlies }`, is it fine, but it should not be a concern at the API level.

fmigneault:
> First we should clarify whether {N} is using base-0 indexing or base-1 (I added the parameter with a description saying it is base 0 indexing).

Agreed. 0-based makes sense.

My bad about `/outputs` | `/results` swap. I did mean the results endpoints as you guessed.

> object would have some very limited utility [...]

The point is that, because different APIs/implementations choose what they consider "sufficiently large" (I don't recall the exact wording in the spec) to pick minimal/representation, you need a way to say you would like the other variant if that is what the client can handle.

Because some APIs might support or not the `../{N}` endpoint, you cannot rely on `{ "href": "/jobs/{jobId}/results/{outputID}/0" }` to work, so you need the server to tell you explicitly which endpoint to use. It might also be the case where the data is not even stored there, so the links must be requested explicitly.

jerstlouis:
@fmigneault 

We're getting quite deep in discussions of Part 3, and I really think for Part 1 it should be a URI, as it was before this PR, and as this PR leaves it.

Regarding the design of Part 3: Workflow execution request, I'm not sure you understood my point about the nested JSON process, so I'll try to illustrate it with an example (though we certainly must have discussed this before in our long workflows-related exchanges):

```json
{
   "process": "https://ecere.ca/ogcapi/processes/ProcessA",
   "inputs":{
      "foo": {
         "process": "https://crim.ca/ogcapi/processes/ProcessB",
         "inputs": {
            "bar": {
               "process": "ProcessC"
               "inputs": {
                  "thing": 123
               }
            }
         }
      }
   }
}
```

My server receives this execution request (`POST https://ecere.ca/ogcapi/processes/ProcessA/execution`), and a very simple implementation of Part 3: Workflows (and a conscious design principle of it) is that I can simply take the object block after `"foo": { ... }` and use it as the payload for the POST operation to `https://crim.ca/ogcapi/processes/ProcessB/execution`, without having to change anything inside:

`POST https://crim.ca/ogcapi/processes/ProcessB/execution`

```json
      {
         "process": "https://crim.ca/ogcapi/processes/ProcessB",
         "inputs": {
            "bar": {
               "process": "ProcessC"
               "inputs": {
                  "thing": 123
               }
            }
         }
      }
```

This is why if we allow a short-hand processID e.g. `"ProcessC"`, it is a processID on your server in this context, and relative URIs cannot be based on the initial server to which the whole workflow was posted. (and I'm realizing that I probably need to change the landing page / `{root}/processes` description to clarify that)

pvretano:
All, be very careful when making changes.  We don't want to make any technical changes unless absolutely necessary and in that case those changes should be broken out in  separate dedicated issues so that we can track the changes and the reasons for them.  Changes to parts that are not part of the RFC (e.g. Part 3 Workflows) are fine though ...

Remember, the goal here is to have everything done by the next SWG meeting so that we can let OGC know that they can start the RFC.  If we manage to do that then there may be an outside chance that we could request a TC adoption vote at the Helsinki TC which would be wonderful.  I'm not holding my breath for that but it is a possibility!

jerstlouis:
@fmigneault 

> Because some APIs might support or not the ../{N} endpoint, you cannot rely on { "href": "/jobs/{jobId}/results/{outputID}/0" } to work, so you need the server to tell you explicitly which endpoint to use. It might also be the case where the data is not even stored there, so the links must be requested explicitly.

If the client wants links, it can do a `GET /jobs/{jobId}/results/{outputID}` with `Prefer: return=minimal`, and it has all the links. Unless the `/{N}` end-point is a requirement (rather than a recommendation), clients cannot actually rely on it at all anyways.

If the client rather wants the actual output, then it can follow the links from `/jobs/{jobId}/results/{outputID}` after requesting it with `return=minimal` (for outputs where maxOccurs > 1), or it can request it with `Prefer: return=representation` for outputs where maxOccurs = 1 (without having to follow any additional link).

fmigneault:
@jerstlouis 
I still don't understand how the resolution would fail from relative URIs.

As you mentioned, the `POST https://ecere.ca/ogcapi/processes/ProcessA/execution` receives that body. So, you know this is the context it applies to. If I use: 

```json
{
   "process": "https://ecere.ca/ogcapi/processes/ProcessA",
   "inputs":{
      "foo": {
         "process": "../../ProcessB",
         "inputs": {
            "bar": {
               "process": "ProcessC"
               "inputs": {
                  "thing": 123
               }
            }
         }
      }
   }
}
```

The `ProcessB` should still resolve to `https://ecere.ca/ogcapi/processes/ProcessB` relatively to `https://ecere.ca/ogcapi/processes/ProcessA/execution` were submitted.

If you then use the nested block, append `./execution` just like you did with the full URI, you still end up with `https://ecere.ca/ogcapi/processes/ProcessB/execution` by making the link absolute. Even if the link was left relative, sending to  `https://ecere.ca/ogcapi/processes/ProcessB/execute`, the `../../ProcessB` would still point to itself (`ProcessB`). 

From my point of view, as @pvretano indicates, this was already possible, so it should not be changed. It is still technically (in terms of schema validation) completely possible. It is just the wording in the description that unnecessarily introduces "absolute" limitation.

Now, why would one want to do `../../ProcessB` rather than `ProcessB` it they resolve the same?
A concrete demo of my previous description is as follows. If the server offers multiple `{root}` (eg: some processes in different namespaces for prod-ready vs in test/dev, or 2 parallel implementations running in separate dockers offered by the server), it could do the following: 

```json
{
   "process": "ProcessA",
   "inputs":{
      "foo": {
         "process": "../../../../ogcapi-testbedXX/processes/ProcessB",
         "inputs": {
            "bar": {
               "process": "../../../../ogcapi-dev/processes/ProcessC"
               "inputs": {
                  "thing": 123
               }
            }
         }
      }
   }
}
```

Then, I can reuse the same workflow across different institutions that share this multi-`{root}` deployment, without defining an institution-specific workflow (i.e.: with `https://ecere.ca/ogcapi/...` explicitly).

---

For the links, you still assume that the ones returned within the array at `/jobs/{jobId}/results/{outputID}` would be the same as the ones in `/jobs/{jobId}/results/{outputID}/{N}`. There can be specific storage limitations that do not allow this, or is not desired to do so. You could also perform content-negotiation on that Nth endpoint to obtain a derived format. Requesting `/jobs/{jobId}/results/{outputID}` with links could trigger an entire staging-out and conversion operation of hundreds of unnecessary files, just because you need the Nth one. That is inefficient if you know just the Nth one needs to be accessed differently to just do stage-out on-demand for that specific entry with the `../{N}` endpoint.

The `/jobs/{jobId}/results/{outputID}/{N}` endpoint is entirely a convenience thing to avoid dealing with the array of `/jobs/{jobId}/results/{outputID}` and address the results individually, as if the process produced them one by one. If it doesn't offer the same capabilities, including `Prefer` return and format selection, it is basically useless!

jerstlouis:
> If you then use the nested block, append ./execution just like you did with the full URI

If a nested process is receiving a collection or process hosted on the same server as that nested process, and you use a collection/process ID only, that would not mean a process or collection on the server to which you are posting the execution request as a whole.

And similarly, if we were to allow relative URI references, it would end up meaning different things when you post it to the server of the top-level process vs. posting the exact same nested block to nested process.

It cannot work.

> From my point of view, as @pvretano indicates, this was already possible, so it should not be changed.

I did not change the Part 1 execute.yaml -- it was already `uri`, not `uri-reference`.

Otherwise this is the Part 3 stuff, where I never considered relative URIs as a valid option because of the above. This might have been changed when migrating this collection input from Part 3 into Part 1, but as I'm pointing out, it cannot work.

---

> If it doesn't offer the same capabilities, including Prefer return and format selection, it is basically useless!

What I think is useless is a `Prefer:` header on `/{N}`, but if you and others think it is useful, I don't mind adding it back.

> Requesting `/jobs/{jobId}/results/{outputID}` with links could trigger an entire staging-out and conversion operation of hundreds of unnecessary files, just because you need the Nth one.

How would you know how many results were produced if you don't first request `/jobs/{jobId}/results/{outputID}`? If the process finished, the staging is probably already done?

If the server supports `/{N}` at all (and the client can only assume it does if this is a requirement), it doesn't need to ask the server to give it a link with a `Prefer: response=minimal` -- `/jobs/{jobId}/results/{outputID}/{N}` *IS* already a valid link (which can support content negotiation).

fmigneault:
> if we were to allow relative URI references, it would end up meaning different things

That is exactly my point. Because it **COULD** mean different things depending on context, it must be supported, so it leaves it open to the client/server to communicate the appropriate thing contextually if that is what needs to be done. Adding relative URI does not disallow passing the absolute URI or the ID explicitly to address their specific contexts. They are just different way to resolve different references nested by the relevant process/collection at the specific moment that it needs to get resolved. Adding the "absolute" restriction unnecessarily limits use cases that can be supported.

---

> How would you know how many results were produced if you don't first request /jobs/{jobId}/results/{outputID}? If the process finished, the staging is probably already done?

Because the process says the minOccurs and succeeded, I do not need to check. It can be assumed that these results would be available in the original format that was generated by the execution.

Depending on how the process was originally executed and its return preference, there might not yet be any staging done (eg: data returned inline), or the staged result might just hold the original format (eg: returned by reference) but not provide the derived product. Doing that staging is another step, and it needs content negotiation for media-type and return preference to obtain what might not have been done by the original execution request.

jerstlouis:
@fmigneault 

> They are just different way to resolve different references nested by the relevant process/collection at the specific moment that it needs to get resolved.

Now I'm confused. Are you suggesting that the relative URI references are relative to execution endpoint to which the specific process object would theoretically be posted, assuming the implementation does this simple "just pass the nested block as-is approach" (which is not a given, as the parent processes could potentially transform that somehow as part of a more complex workflow orchestration / caching etc.)?

If so, yes that could work in theory, but that means that some of the relative URIs (that are actually ignored by the top-level process living on that first server to which the execution request is posted), are resolved to a different URI than the one that the workflow as a whole is posted. i.e., they work just like the process / collectionID. That would be rather confusing in my opinion, but if that is what you mean, we could potentially consider that as at least it doesn't break the whole concept.

---

> Because the process says the minOccurs and succeeded, I do not need to check.

But minOccurs is likely usually 0 or 1... Sure, you can check `/0` if it's 1, but for the additional values you need to check, or somehow loop until you get a 404.

> and it needs content negotiation for media-type and return preference to obtain what might not have been done by the original execution request.

Sure it needs content negotiation for media-type and `/{N}` can support that. But I still don't understand why it would need a `Prefer: return=` at that end-point? Why can't it always return the output itself (`Prefer: return=representation`)?
What is the use of `Prefer: return=minimal` there? Given that a client that needs a link can always use a pseudo-response `{ "href": "/jobs/{jobsID}/results/{outputID}/{N}" }`? And if it's because the actual data sits on another server, that's not an issue either since the server can simply redirect to wherever the actual data is (taking into consideration the `Accept:` header).

fmigneault:
I would like the standard to be really clear about which interpretation it should be. There is a mention that it must be relative to the nested process, but that is still not clear enough "when" it happens. 

I think it makes sense if the "../" paths are passed as is to the nested process, and then it resolves during its own execute request context where that refers to. It just has to be clarified, and an example would go a long way to demonstrate the distinction. 

>  they work just like the process / collectionID

Yes, it would essentially be like this, but not limited to the same `{root}` of processes since you could rewind the source further to another process catalogue. It also helps support cases where collections and processes do not even reside under the same `{root}` (eg: they don't share the same `ogcapi/` prefix). 

I think it is fair that, if a URI should be aligned across multiple nested processes, it should be pre-resolved (ie: before sending the workflow to the server) as absolute URL since that is the only way to make sure they all point at the same place no matter where nested parts are sent.


---

The main case I expect is that the server responds with an array of `href`, so I already know there are many results. I don't need to double-check them. Then, the user wants to look one up to verify it, but the original data is not directly viewable (eg: a NetCDF or whatever) so some kind of PNG is needed, and `Prefer` allows getting the data preview rather than another link. The storage does not itself offer conversion, but the OGC API does, so it either needs another stage-out operation for that PNG representation, or return the data directly.  Of course, the results could use `{ "href": "/jobs/{jobsID}/results/{outputID}/{N}" }`, but if that endpoint also returns `href` because the API decides it is too big to return the data blob, I'm just turning in circle.

Basically, if I even bother implementing the `../results/{outputID}/{N}` endpoint with content-negotiation, so it can work just like a `../results/{outputID}` with single-output, I will literally use the same code, so it will by default support all the same prefer representation convenience. How the client actually interacts with the `./{N}` endpoint to get a link or data is up to it, for whichever use case it needs to fulfill.

jerstlouis:
@fmigneault 

> I think it is fair that, if a URI should be aligned across multiple nested processes, it should be pre-resolved (ie: before sending the workflow to the server) as absolute URL since that is the only way to make sure they all point at the same place no matter where nested parts are sent.

The need for pre-resolving is exactly what I am very strongly opposed to. It prevents a simple implementation and the elegance of the nested workflows. And since not all relative URIs in an execution request would be relative to the same base URI, it would cause confusion (that could be clarified with clear language and examples). But I would still much prefer that confusion over breaking the design goal of being able to send a nested block without modification to a nested process, if you really want those relative URI for your use case of referencing another API sitting on the same domain but outside the landing page (or you could simply use the full URI for first top-level process for each different OGC API servers -- you would still only have a single endpoint to replace per different API, and you use processID and collectionID for the deeper ones on that same server, and you don't have to wonder whether you got the number of `../` right).

And if this programmatic, you can also simply use a `{template}` for that shared domain that you replace before submitting the execution workflow.

---

>  and Prefer allows getting the data preview rather than another link. 

That is a different use case from `Prefer: return=minimal` vs. `Prefer: return=representation`. If you mean a proper representation in PNG, then the content negotiation would support that. But if you mean a preview that is not the actual data at all, I don't think the standard defines anything for that? There is no `Prefer: return=preview`, is there?

> Of course, the results could use `{ "href": "/jobs/{jobsID}/results/{outputID}/{N}" }`, but if that endpoint also returns href because the API decides it is too big to return the data blob, I'm just turning in circle.

I'm not sure I understand what you meant there. `/jobs/{jobsID}/results/{outputID}/{N}` would **never** returns an `href` (`Prefer: return=minimal`)... It would **always** return the data itself (`Prefer: return=representation`). That is my entire argument why it is unnecessary to support `Prefer:` at this endpoint.

fmigneault:
Maybe I wasn't clear enough. I think you misinterpret my comment. 

 By "pre-resolving", I didn't mean that the OAP server does it. The _**client**_ that submits the request pre-resolves anything it needs to as absolute URLs *if that is what they want*. If they don't (ie: either relative URI or plain IDs are left as is in the submitted workflow), then OAP will resolve them in-time, contextually and individually for each block, where each of those nested block is sent and executed.  

To me, there is no distinction between relative URIs and plain IDs, beside the `{root}` they refer to. They are just as confusing in the sense that they apply contextually to the nested block, not the "top-most/main" process execution endpoint. That means that relative URIs and IDs can "change context" midway depending on which nesting level we are processing, and that makes reading the workflow payload confusing. But conceptually, they are both "relative". The only way to have the same "main" `{root}` across nested blocks is to provide them as absolute URIs. There is no breakage in the design.

fmigneault:
Maybe these concrete deployment examples can help you better understand my concerns...

- CRIM Node:
  - https://hirondelle.crim.ca/weaver/
  - https://hirondelle.crim.ca/stac/
  - https://hirondelle.crim.ca/geoserver/ogc/coverages/v1
  - ...
- UofT Node: 
  - https://redoak.cs.toronto.edu/weaver/
  - https://redoak.cs.toronto.edu/stac/
  - https://redoak.cs.toronto.edu/geoserver/ogc/coverages/v1
  - ...
- Ouranos Node:
  - https://pavics.ouranos.ca/weaver/
  - https://pavics.ouranos.ca/stac/
  - https://pavics.ouranos.ca/geoserver/ogc/dggs/v1
  - ...
- PCIC Node:
  - https://garryoak.pacificclimate.org/weaver/
  - https://garryoak.pacificclimate.org/stac/
  - https://garryoak.pacificclimate.org/geoserver/ogc/coverages/v1
  - ...
- ...

None of our nodes share a common "ogcapi" `{root}` across all OGC services they offer, so I cannot throw in a random `{processID}` or `{collectionID}` without context. They won't know which service to resolve if I do. We need to be able to dispatch workflows *dynamically* and *across nodes*. If I hardcode URIs, I just have a bunch of duplicate workflow deployments per node. If I can use relative URIs however, I can share equivalent workflows across all nodes, even though services don't share a `{root}` since each node can still resolve its processes/collections relatively to where the request was sent, in-time and contextually when needed.

That is why both absolute URIs and plain IDs to not resolve my use case. Only relative URIs do.

jerstlouis:
@fmigneault 

> then OAP will resolve them in-time, contextually and individually for each block, where each of those nested block is sent and executed.

OK, then I think that is the _not all relative URIs in an execution request would be relative to the same base URI, it would cause confusion (that could be clarified with clear language and examples)_ scenario I was talking about. Which I much prefer over all relative URIs are submitted to where the whole execution request is first posted to.

> They are just as confusing in the sense that they apply contextually to the nested block, not the "top-most/main" process execution endpoint. 

For me they are significantly more confusing, first because they are relative to URIs that do not appear anywhere in the workflow (`.../execution`), and in this nested context, not relative to the original URI to which the workflow is initially posted.

But if everyone's OK with these URI references that are relative to the execution endpoint of the process receiving a nested process value or collection value object, I can make the changes in the OpenAPI definition, and @pvretano can review the Collection Input. Note that I think this would not have an impact on the top-level process for Part 1, which should always be absolute or is otherwise useless. The uri-reference for `{ "process": ... }` would only be an option for Part 3, and collection value would also accept a uri-reference.

>  But conceptually, they are both "relative".

Yes you can think of it that way, but using a processId or collectionId you don't need to think about how many `../` are needed :)

jerstlouis:
@fmigneault 

> Maybe these concrete deployment examples can help you better understand my concerns...

I understood your use case from the start, but I would use either of these two solutions for it:
- Unify these on a single OGC API root (I'm envisioning a single global federated OGC API, but you're splitting your own server into pieces :)). A simple way to do this would be to add the collections from STAC and GeoServer to `/weaver/collections/` (the access mechanisms links there can be relative links to `../../STAC` etc.). You could add a capability to "register collections" from another API (whether on the same domain or a remote domain)
- Use a `{serverURL}` template in the workflow definition that you search & replace programmatically

But like I said in https://github.com/opengeospatial/ogcapi-processes/pull/568#issuecomment-4180125054 , as long as it's relative to the receiving process server, I'm okay with adding a uri-reference. We just need to explain very clearly what it's relative to (`/processes/{processId}/execution` of the process receiving the input value?)

So the relative uri-reference way to point to another process on the same server as the process receiving the nested process would be `../otherProcess` ?

jerstlouis:
@fmigneault 

In your examples before you used `../../ProcessB` and `../processes/other`.

But per [RFC3986 Section 5.1](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1):

>  If the base URI is obtained from a URI reference, then that reference must be converted to absolute form and stripped of any fragment component prior to its use as a base URI.

when adding this to `/processes/ProcessA/execution` it first gets stripped to:

`/processes/Process/`

and so that would yield

`/ProcessB` and `/processes/processes/other`

So I'm guessing you meant:

`../ProcessB` and `../../processes/other`

aren't relative URIs much more confusing than a processId? :)

A large part of the confusion is the base URI for the relative uri-references in the document are not all to the top-level execution end-point, and don't appear in the document either. And I'm actually not sure whether what we're defining reasonably fits within the 4 allowed ways in section 5.1:

- [5.1.1](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.1).  Base URI Embedded in Content (base URI does not appear anywhere)
- [5.1.2](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.2).  Base URI from the Encapsulating Entity (perhaps we can consider the receiving process as an "Encapsulating Entity"?)
- [5.1.3](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.3).  Base URI from the Retrieval URI (I'm assuming this extends to the POST URI and works for the top-level process...)
- [5.1.4](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1.4).  Default Base URI (does not apply either -- the base URI differs for different uri-references)

Another issue is that feeding the resolution of `clip` to a `buffer` process interpreted as a uri-reference vs. a {processId} differs if we consider the `/processes/buffer/execution` end-point of the receiving process rather than the process itself (`/processes/buffer`):

- processID says it's `/processes/clip`
- uri-reference says it's `/processes/buffer/clip`

So we would need a special rule that if it's a string id, then it's `/processes/{processId}` that overrides resolving it as a URI reference.

So perhaps the best thing is actually to define it as relative to the `process URI` of the receiving process that appears in the execution request. I think that makes a lot more sense actually. So it is not about where the nested process object will be posted. The top-level process of a workflow being submitted is never allowed to be a uri-reference: either you omit it/discard it, or you turn it into a processID or an absolute URI that matches what you're post it to (or we can simply say that the server can ignore it when posted to `/processes/{processId}/execution`).

So I think this is the solution... The receiving process URI (not the `.../execution`) endpoint is the base URI, based on the section 5.1.2 "Encapsulating Entity" concept. Although we still need to make an exception for a collectionId which are at `../collections/{collectiondId}`, so that if no `/` is used in the `collection` this should be interpreted as a collectionId rather than a relative URI reference.

jerstlouis:
@fmigneault @pvretano 

Please review the revised changes (for `collection-value.yaml` and `execute-workflows.yaml`).

I believe this is the best possible way to reconcile:
1. Francis's use case of multiple microservices on the same domain, while re-using the same identical execution request on different domains
2. The simple workflow implementation use case of simply passing down nested process execution blocks verbatim to nested processes
3. [RFC 3986 Section 5.1](https://datatracker.ietf.org/doc/html/rfc3986#section-5.1)

In short:

- Relative URIs are relative to the process description (not the `/execution` endpoint)
- At the top-level, only absolute URI should be used for `process` (which allows GDAL to open an execution request as a data source), but the server must ignore it to not break 2 (except for a potential Part 4 with POST to `/jobs`)
- For collection outputs, the base URI to resolve a relative collection URI reference would depend on which resource is retrieved (`/jobs/{jobID}/results`, `/jobs/{jobID}/results/{outputID}`, or `/jobs/{jobID}/results/{outputID}/{N}`)
- For collection inputs/outputs, a collection ID without a `/` must be resolved as `{root}/collections/{collectionId}`, where `{root}` is the landing page of the receiving process / retrieved output document

pvretano (approved):

fmigneault:
> * Unify these on a single OGC API root (I'm envisioning a single global federated OGC API, but you're splitting your own server into pieces :))

Unfortunately, I cannot do that. Different institutions manage different components per application domains. I think this kind of scenario is fairly common, whether managed by different departments, countries, provinces, etc., so if the API was not robust enough to support this kind of use case and single/federated API was always assumed, I would see it as a big red flag in terms of design. I think the latest proposal allows the flexibility. 

> Use a {serverURL} template in the workflow definition that you search & replace programmatically

Yes, that can work to "patch it", but then I cannot expose/publish the workflow externally for users to reuse it unless they all apply the same patches. Relative URIs are the "good enough" portable solution. 

> So perhaps the best thing is actually to define it as relative to the process URI of the receiving process that appears in the execution request. I think that makes a lot more sense actually.

Agreed. Since the process could also be executed with `POST /jobs` of Part 4, the execution endpoint makes little sense. The only thing consistent is the process URI itself.

I am still worried though about the `{processID}` form. I don't if there could be a way to make the relative context more obvious from its value alone? It still seems very easy to mistake for another location since one always need to look one level higher to figure out what it actually refers to. So far, all our examples are fairly simple, but I can imagine workflows becoming hard to read when there are a lot of input/outputs and nested processes involved.

The current updates a good at least to support the functionalities assuming users read the docs properly.

openapi/schemas/processes-workflows/execute-workflows.yaml
```diff
@@ -4,7 +4,18 @@ allOf:
         process:
           type: string
           format: uri-reference
-          description: URI to the process description end point (i.e., `.../processes/{processId}`)
+          description: |-
```
fmigneault, line 7: Can use `>-` to remove the extra `\n`. Just make sure to use double newlines for actual paragraphs/list-items to preserve as separate blocks.

jerstlouis:
@fmigneault 

> I am still worried though about the {processID} form.

You don't need to worry about this one, because with the base URI being the process description, the `{processID}` form is fully equivalent to relative URI reference consisting of the process ID -- it's no longer a "separate form".

The only special handling that remains is for the `{collectionID}` for collection inputs.

Those need to be differentiated from a relative URI, based on the fact that they contain no `/`, so that `countries` resolves to `/collections/countries` rather than `/processes/countries`.

But whether resolving a relative URI or an identifier, you need to look at the same process receiving the input in both cases.

openapi/schemas/processes-workflows/execute-workflows.yaml
jerstlouis: @fmigneault my YAML skills are horrible. Can we also just write :

```yaml
   description: |- URI to the process description end-point
      (i.e., ...
```
to avoid the newline?

fmigneault: The `|` preserves `\n`. `>` removes them unless they are double (to make a paragraph). The `-` is just for the final one.

fmigneault:
> You don't need to worry about this one, because with the base URI being the process description, the `{processID}` form is fully equivalent to relative URI reference consisting of the process ID -- it's no longer a "separate form".

The case I could see where it becomes complex to interpret is if there are multiple nested processes that switch context with intermediate absolute URL. From a functional standpoint, it still makes sense, but `B`, `D`, `E`, etc. can be hard to track (for a user reading the workflow) with many inputs in between. Notably, if `E` was much lower in nesting, finding `C` as the parent context can be tricky (for a user, not the engine) (ie: how many levels must be rewinded to find the `{root}` they refer to).

```yaml
process: https://server1.example.com/processes/A
inputs:
  in: 
    process: B
    inputs: 
      in: 
        process: https://server2.example.com/processes/C
        inputs:
          in:
            process: D
            inputs:
              in:
                process: E
                inputs: {...}
```

jerstlouis:
> Notably, if E was much lower in nesting, finding C as the parent context can be tricky (for a user, not the engine)

While the JSON notation is simple enough that it is quite easy to edit manually, users can of course use a workflow editor to edit a visual workflow graph. But otherwise you just go back up until you see an absolute URL :)

Issue #563, opened 2026-03-19
DRU: rWorkflowNotFound.yaml was never added


jerstlouis:
The file rWorkflowNotFound.yaml is included by:

- paths/processes-dru/pDeploy.yaml
- paths/processes-dru/pProcessDescriptionReplaceUndeploy.yaml
- paths/processes-dru/pProcessListDeploy.yaml
- responses/processes-dru/rWorkflowNotFound.yaml
- paths/processes-dru/pReplace.yaml

but it was not added together with these changes in 3cd54af60cd7fede3fc9647aa39c5fbeddd219f0

I added a guess of what this file could be in #561 ( https://github.com/opengeospatial/ogcapi-processes/blob/8a69bf6c20adc83da42fad4a1674973b98eb3e26/openapi/responses/processes-dru/rWorkflowNotFound.yaml ), but @gfenoy you might have some specific content for it.

cc. @pvretano

jerstlouis:
@gfenoy Could you please compare the provisional [rWorkflowNotFound.yaml](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/responses/processes-dru/rWorkflowNotFound.yaml) with your version that you forgot to add in 3cd54af60cd7fede3fc9647aa39c5fbeddd219f0, and submit a PR with any changes needed?

Thanks!

jerstlouis:
Closing as the PR fixing this has been merged.

PR #569, opened 2026-03-31
Add the missing rWorkflowNotFound.yaml

gfenoy:
Following issue #563, we found the missing file.

We also include the [ogcapi-processes.yaml](https://github.com/opengeospatial/ogcapi-processes/compare/master...GeoLabs:ogcapi-processes:feature/fix-issue-464?expand=1#diff-524ebf61d9e0967bd8c9692a2eae07dad5b3dbfcbb40d4fa95c63a02af610267) and the bundled version.


jerstlouis:
Thank you very much @gfenoy for looking for and finding the missing file!

The main difference with the current version on `master` is:

```diff
- description: The requested workflow does not exist on the server.
+ description: workflow was not found in the CWL
```
Looking closer at the paths using this response, it is actually used for:
- the `400` response of the deploy operation (POST to `/processes`)
- the `400` response of the replace operation (PUT `/processes/{processId}`)

EDITED:

Could we also remove the `pReplace.yaml`, `pDeploy.yaml`, `pUndeploy.yaml`, and just have the single source of truth  in `pProcessDescriptionReplaceUndeploy.yaml`, instead of relying on bash commands to concatenate them together? This `paths/` directory is intended to be one file per path, not one file per operation.

It seems that neither of these descriptions are actually adequate, as it is not specific to CWL at all.

Could we simply use [`rException.yaml`](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/responses/common-core/rException.yaml), so that any details about the error can be provided to the client?

(I also note that your PR was not based on the latest master branch, but 170 commits behind)

Thank you!

gfenoy:
@jerstlouis thank you for looking into this PR. Indeed, the file we found was created back in 2024 and should have been included in this branch at that time. Sorry for this mistake.

At the time we added this file, it was only to give a way to select a workflow identifier included in a CWL graph. In consequence, I would be in favor of the wording you provided: "The application package being deployed is invalid.". 

Also, the advantage of having a separate file for each path is that a server implementation can use the schemas only for the paths they support. So, although it is not used internally, Server Implementation can rely on these schemas. Also, I think it makes sense to keep schemas small and aggregate them when needed, rather than adding a full-featured path like [pProcessDescriptionReplaceUndeploy.yaml ](https://github.com/opengeospatial/ogcapi-processes/blob/v2-openapi-fixes/openapi/paths/processes-dru/pProcessDescriptionReplaceUndeploy.yaml). We may investigate ways to create this file dynamically based on the existing schemas or by aggregating them.

It is ok to close this PR, and you can prepare a new one. Nevertheless, in the current Part 2, there is no `404` exception defined; we are re-using the one defined in Part 1, which is exactly what we need. I don't see the interrest to add the "or package" to the exception.

When a client reaches `404`, it typically means that the `processID` does not exist. I am not sure adding an exception for 404 when we have one in Part 1 makes much sense.

I wait for your feedback to close this PR.

jerstlouis:
Thank you @gfenoy . Sorry I was mixed up in my response above -- I edited it.

You are correct this was not used for a 404, I was confused.

> Also, I think it makes sense to keep schemas small and aggregate them when needed, rather than adding a full-featured path like [pProcessDescriptionReplaceUndeploy.yaml ](https://github.com/opengeospatial/ogcapi-processes/blob/v2-openapi-fixes/openapi/paths/processes-dru/pProcessDescriptionReplaceUndeploy.yaml). We may investigate ways to create this file dynamically based on the existing schemas or by aggregating them.

This was largely the source of my confusion (too many files).
These are different operations for the same paths.
The original intent with this structure was to have one file per **path**, not operation.

The one advantage I see with the `update.sh` approach is to synchronize the Core paths with the DRU operations.
I think what would help remove the confusion is to define them as `paths/processes-dru/operations/oDeploy.yaml`, `paths/processes-dru/operations/oReplace.yaml`, `paths/processes-dru/operations/oUndeploy.yaml`. This way we would not be mixed up thinking that they are actual paths that are not used...

Regarding the 400, should we use [`rException.yaml`](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/responses/common-core/rException.yaml) for the 400 case? This way the server can provide a more detailed exception report about why the application package failed to deploy?

I'll do a PR for you to consider. Thanks!

jerstlouis:
@gfenoy Please see #570 as my proposed alternate fix for this issue.

It keeps the bash-assisted combination but makes it a bit more tidy / less confusing by moving them into a sub-directory for operations, to avoid confusion with the actual path resources. And uses `rException.yaml` for the 400.

jerstlouis:
We'll use the generic rException.yaml so that server can provide additional details as to why the application package deployment file, so #570 is the proposed fix for #563.

Thank you and sorry for the extra efforts @gfenoy .

PR #570, opened 2026-03-31
Proposed fix for #563

jerstlouis:
Proposed fixes for #563

- keep the bash-assisted combination, but make it a bit more tidy / less confusing by moving the individual path operations into a sub-directory for operations, avoiding confusion with the actual path resources
- use `rException.yaml` for the 400 so server can provide a detailed response to the client about what went wrong with the deployment


openapi/update.sh
gfenoy, line 1: I like a lot the idea of adding an operations subdirectory.

gfenoy (approved):
I like a lot the idea of adding an operations subdirectory.

pvretano (approved):

PR #579, opened 2026-04-15
Openapi fixes

jerstlouis:
This fixes new OpenAPI errors and issue #572.

PR #581, opened 2026-04-27
V2 openapi fixes

fmigneault:
>[!NOTE]
> Following is the same as #582, but *without* the extra plugin that will apply the standard's name as prefix (eg: subdirs in https://github.com/opengeospatial/ogcapi-processes/tree/master/openapi/schemas). It is the simplest `redocly` integration, which introduces `-#` suffixes conflict names in the generated bundle.
> 
> - Closes #582 (alternative)

- Includes @jerstlouis PR #579  and #565 
- Includes a fix of incorrect reference to `CWLWorkflowStepOut`


fmigneault:
@jerstlouis @pvretano Ready to review.

The main diffs between `redocly` and `swagger-cli` are:

- `redocly` prefers using the direct schemas, so the "redundant/redirect" `$ref` components are not used
  e.g.: `"$ref": "#/components/responses/rLandingPage"`  is used directly in the response rather than pointing at `"$ref": "#/components/responses/LandingPage"` (see no `r`), with the extra `LandingPage => rLandingPage` redundant component. These extra components remain there, but they are superfluous (only `swagger-cli` needs them explicitly).
  
- `redocly` relies on the schema names, so "conflicts" like duplicate use of `"link.yaml"` at different subdir locations (https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/common-core/link.yaml and https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-core/link.yaml) generates `link` and `link-2` names in the bundle to avoid the flat-list conflict
  - see cases flagged in "*Bundle OpenAPI*" logs under https://github.com/opengeospatial/ogcapi-processes/actions/runs/25004811067/job/73224530986?pr=581
  - This is a known issue (https://github.com/Redocly/redocly-cli/issues/661) and would require a custom handler if we want to apply a specific naming strategy (e.g.: prefix by the OGC standard that contains each schema)

fmigneault:
@pvretano @jerstlouis Up to date with master. Awaiting review.

PR #582, opened 2026-04-27
Redocly plugin

fmigneault:
>[!NOTE]
> Following is the same as #581, but with an extra plugin that will apply the standard's name as prefix (eg: subdirs in https://github.com/opengeospatial/ogcapi-processes/tree/master/openapi/schemas) before the individual schema file names to ensure they have unique, which does not require `redocly` to apply a `-#` suffix.
> 
> - Closes #581 (alternative)

FYI @jerstlouis @pvretano 

I let you decide which of the 2 approaches is preferred.
IMO, this one is the cleanest because it makes it clear about the standard's source for each schema. 
Either way, the linter works with the final result to test if the bundle is valid altogether.


fmigneault:
@pvretano @jerstlouis Up to date with master. Awaiting review.

PR #591, opened 2026-06-22
Alignment with Features Part 4.  Fix execution unit.

### LINKED ISSUE #397: The execution unit schema is really confusing!
I have been reviewing the Part 2 and the schema for the execution unit is really disjoint and confusing.  

If the execution unit is a container then the execution unit looks like this:

````
"executionUnit": {
  "type": "docker",
  "image": "mydocker/ndvi:latest",
  "deployment": "local",
  "config": {
    "cpuMin": 2,
    "cpuMax": 5,
    "memoryMin": 1,
    "memoryMax": 3
  }
}
````

If, however, the execution unit is CWL, the execution unit looks like this:

````
  "executionUnit": {
    "href": "https://raw.githubusercontent.com/EOEPCA/app-snuggs/main/app-package.cwl",
    "type": "application/cwl"
  }
````

or like this depending on whether the CWL is inline or referenced:

````
{
  "executionUnit": {
    "value": { ... },
    "mediaType": "application/cwl+json"
  }
}
````

Why such varying schemas?  Why, for example, does the CWL class have both "type" and "mediaType" when the examples are both media types?  Why not have a single member, `type`, of type string and simply have a requirement in the CWL conformance class that it's value has to be a media type?

Also, why does the docker conformance class have a member `image`?  Why not call it `value`  to be consistent and simply include a requirement is the docker conformance class that the value of `value` is a Docker/OCI image reference?  The docker
conformance class could further extend a "base" schema to add the `deployment` and `config` members.

If, for example, the base schema of the execution unit was:

````
type: object
required:
  - type
  - value
properties:
  type:
    type: string
  value:
    type: string
`````

The docker conformance class could extend this schema like this:

````
allOf:
  - $ref: executionUnit.yaml
  - type: object
    properties:
      type:
        type: string
        unum:
          - docker
          - oci
      deployment:
        type: string
        enum:
          - local
          - remote
          - hpc
          - cloud
      config:
        type: object
        properties:
          cpuMin:
            type: number
            minimum: 1
          cpuMax:
            type: number
          memoryMin:
            type: number
          memoryMax:
            type: number
          storageTempMin:
            type: number
          storageOutputsMin:
            type: number
          jobTimeout:
            type: number
````

And the CWL conformance class could extend this schema like this:

````
allOf:
  - $ref: executionUnit.yaml
  - type: object
    properties:
      value:
        oneOf:
          - type: string
          - type: object
````

... and, of course, a vendor-specific execution unit such as:

````
"executionUnit": {
  "type": "python",
  "value": "... my python program text ..."
}
````

is also schema valid allowing experimention to flourish without too much trouble and perhaps lead to additional execution unit conformance classes.

### PR DESCRIPTION
**Pull request by pvretano**
Closes #397

### CODE REVIEW

File: extensions/deploy_replace_undeploy/standard/sections/clause_9_cwl.adoc [hunk 1]
```diff
@@ -65,8 +70,9 @@ In these cases, the <<rc_cwl,CWL>> additional encoding (i.e.: JSON or YAML) need
 ----
 {
   "executionUnit": {
-    "value": "cwlVersion: 'v1.2'\nclass: CommandLineTool\n...",
-    "mediaType": "application/cwl+yaml"
+    "type": "cwl",
+    "mediaType": "application/cwl+yaml",
+    "value": "cwlVersion: 'v1.2'\nclass: CommandLineTool\n..."
   }
```
fmigneault: These definitions were already aligned in a similar form to `qualifiedValue.yaml` vs `link.yaml`, directly provided within the `executionUnit` object, using the usual `value` vs `href` fields.

If desired, the `executionUnit.yaml` (for the `type`/`config`/`image` properties) could be aligned to use the `qualifiedValue.yaml` form, but I believe the CWL examples are already adequately formatted without adding `type` or `link` nesting.

fmigneault (changes requested):
PUT edits are OK. Just the execution unit to review.

extensions/deploy_replace_undeploy/standard/sections/clause_9_cwl.adoc
pvretano: I am confused!  

Issue #397 contains a proposed schema that accommodates almost all the forms of CWL illustrated in clause 9 but does not rely on `qualifiedValue.yaml` (although it seems to be inspired by it). The proposed schema at the end of issue #397 does not include a `mediaType` property  which the CWL examples seem to rely on so should I ...

- extend `qualifiedValue.yaml` to define the schema of the execution unit **OR** 
- should I simplify the proposed schema  in issue #397 and add a `mediaType` property **OR** 
- should I just simply the proposed schema with no `mediaType` property (which would require that the CWL examples in clause 9 be updated to the `type` rather than `mediaType`)?

My preference is to simply the schema proposed in issue #397 to be this:
````
oneOf:
  - $ref: link.yaml
  - anyOf:
    - description: |-
        The execution unit is a DOCKER or OCI container.
      type: object
      properties:
        type:
          enum:
          - docker
          - oci
        value:
          $ref: executionUnitContainer.yaml
      additionalProperties: false
    - description: |-
        The execution unit is a CWL script.
      type: object
      properties:
        type:
          enum:
            - cwl
        value:
            $ref: "https://raw.githubusercontent.com/common-workflow-language/cwl-v1.2/main/json-schema/cwl.yaml"
      additionalProperties: false
    - description: |-
        Other forms of execution unit (e.g. Python script, shell script) are allowed
        by the Standard but are not described in the Standard.
      type: object
      properties:
        type:
          type: string
        value:
          type: object
      additionalProperties: true
````
If I were to define the execution unit schema based on the `qualifiedValue.yaml` schema it might look like this:
````
oneOf:
  - $ref: link.yaml
  - allOf:
    - $ref: qualifiedValue.yaml
    - type: object
      properties:
        value:
          anyOf:
            - type: string
            - type: object
            - $ref: executionUnitContainer.yaml
            - $ref: "https://raw.githubusercontent.com/common-workflow-language/cwl-v1.2/main/json-schema/cwl.yaml"
````
There is nothing explicit in this case telling you that the execution unit is a Docker container however that information could be determined by looking at the `value` object which would include the `image` property. If we wanted to make the Docker case explicit we could add an optional `type` member with the values "docker" or "oci".  There is one other possibility too ... there is no official media type for a docker image.  However, there are media types for the parts of a docker image.  Specifically the media type `application/vnd.oci.image.manifest.v1+json`, which is the media type of the container image manifest, could be used as the value of the `mediaType` property to indicate that the execution unit is a container image. 

@fmigneault, @gfenoy thoughts?

Comment (fmigneault) [reply extensions/deploy_replace_undeploy/standard/sections/clause_9_cwl.adoc]: I think we need a cherry-picked mix of the two variant with explicit types required (one way or another). One thing that is "risky" with plain `type` is that users could come up with their own variants that might not be reliable, when more explicit and reliable `mediaType` applies to cases like CWL. For example, `mediaType: application/cwl+yaml` can indicate more specifically the string contents than `type: cwl`. Indeed, the [OGC/docker media-types](https://github.com/opencontainers/image-spec/blob/main/media-types.md) are very tricky. An optional `type` seems reasonable for them. 

Therefore, something like this?

```yaml
oneOf:
  # 'type' as plain string vs media-type is distinguished via the presence of 'href' or not
  - $ref: link.yaml
  - allOf:
    # separate case for type/mediaType to enforce them to be there with any value
    # actual value combinations defined by enum in the other 'allOf.anyOf' items
    - anyOf:
      - type: object
        required: [type]
        properties: {type: {type: string}}
      - type: object
        required: [mediaType]
        properties: {mediaType: {type: string}}
    - anyOf: 
      # equivalent to value/mediaType, but 'mediaType/type' enforced above
      # this applies on top of the 'value' it already require itself
      # no need anymore for the generic open-object at the end of the 1st example
      - $ref: qualifiedValue.yaml  
      # ...  rest of 1st example, with added mediaType properties
```

pvretano: @fmigneault I had to get Part 2 ready for RFC because OGC was waiting on me so I settled on the schema [here](https://github.com/opengeospatial/ogcapi-processes/blob/master/openapi/schemas/processes-dru/executionUnit.yaml).  Take a look and let me know your thoughts... If you want to make changes, create a new issue and we can deal with is as an RFC comment.

fmigneault: OK looks good. Similar result.

I notice the `executionUnitBase.yaml` contains `required: type / mediaType`. I think they must be arrays.

PR #605, opened 2026-07-09
Part5 Provenance

### LINKED ISSUE #602: Part 4: The provenance conformance class should be split out into its own document
After reviewing Part 4, I think that the provenance part should be split out into its own document (Part 5 I guess).   This information is metadata about a job and I think it would be more clearly described in its own specification.  Also other groups such as the Provenance SWG/DWG might find it easier to comment on rather than it being buried inside of Part 4.

fmigneault:
- First pass of moving Part 4 `Provenance` class into a dedicated Part 5 (closes #602)
- Adds multiple recommendations regarding Provenance contents and how they relate to Processes concepts
- Provide PROV examples with most of the well-established formats, based on `EchoProcess` execution


fmigneault:
@pvretano @bpross-52n @gfenoy 
Still cannot assign myself or set labels. Let me known what you think.

gfenoy:
As we all agreed on this task during the SWG meeting, I can merge the PR.

@fmigneault, please let me know when you want me to proceed.

fmigneault:
@gfenoy I'd like to gather feedback first since it is easier to analyse the diff here. If others are OK with it, then we can merge it.

pvretano:
@fmigneault @gfenoy I've looked it over an it looks fine to me.  I would go ahead and merge and also set up the nightly build of the document. 

There is one other housekeeping thing which I think we need to deal with and that is the charter.  We need to review and revise the charter to reflect the work that is going on in the SWG.  It's not a good idea to let the charter get too far out of sync with the work being done.

PR #614, opened 2026-08-08
fix redocly duplicate refs

fmigneault:
## Changes

- replaced the explicit CWL schema URI in *Part 3: Workflows & Chaining* causing duplicate CWL definitions from local schema used by *Part 2: DRU*
- improved the redocly plugin logic to properly include all CWL subschemas
- added `title` injection (if not already provided) to nicely render the schemas in Swagger-UI (such that the bundle's dir-prefix to manage schema-name conflicts is not shown)

Recent update to redocly slightly tweaked the behaviour (https://github.com/Redocly/redocly-cli/releases/tag/%40redocly%2Fcli%402.45.0), which causes the auto-update bundle diffs seen in other active PRs.

The result is now properly obtained as shown below.

## Result

image

### DIFF
--- a/.github/redocly/schema-name-prefix.plugin.js
+++ b/.github/redocly/schema-name-prefix.plugin.js
@@ -9,6 +9,24 @@
 
 // Store mapping of component objects to their source information
 const componentSourceMap = new WeakMap();
+const externalDefsRefMap = new WeakMap();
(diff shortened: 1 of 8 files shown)


pvretano (approved):
Looks good to me ... go ahead and merge.

PR #626, opened 2026-09-08
Add the missing rPackage response

gfenoy:
Add the missing rPackage response in processes-dru.


fmigneault:
@gfenoy 
Please revert the removal of `application/cwl`. This is a well-known and registered media-type for CWL under https://www.iana.org/assignments/media-types
It should also be added to the new rPackage endpoint for consistency.

